PR #3, opened 2022-05-06
Add Ellipsis Finance adapter

mashup-purveyor:


slasher125:
Hey sir, ty for the adaptor!

a few notes:
- for project (use the defillama slug, which would be `ellipsis-finance` . Also use that for your project folder name
- for chain: `utils.formatChain('binance')`
- u don't have to round the apy value on the adaptor level, we handle that for you on the frontend so please remove that part. same for tvl, no need to round it

mashup-purveyor:
Hey thanks; all done.

mashup-purveyor:
Hey @slasher125,

Thanks for this. On the listings it isn't very apparent how to find these pools on the underlying dApp. For example, in the readme is states for `symbol` that we put, "symbol of the tokens in pool". So we did this, however without a pool `name` it is very hard for the user to go from an Ellipsis entry on the table and find that on our site.

I wonder if it wouldn't be useful to provide a URL to the pool, or at minimum a pool name?

PR #4, opened 2022-05-08
pickle

waynebruce0x:


slasher125:
a small thing i noticed, the contract address used for `pool` is unfortunately not unique. (there are 6dupes). for example:

```
> data.filter(el => el.pool === '0x55D5BCEf2BFD4921B8790525FF87919c2E26bD03')
[
  {
    pool: '0x55D5BCEf2BFD4921B8790525FF87919c2E26bD03',
    chain: 'Moonriver',
    project: 'pickle',
    symbol: 'SOLARSOLAR-MOVR',
    tvlUsd: 430.301,
    apy: 1.881
  },
  {
    pool: '0x55D5BCEf2BFD4921B8790525FF87919c2E26bD03',
    chain: 'Cronos',
    project: 'pickle',
    symbol: 'VVSCRO-ETH',
    tvlUsd: 5.52,
    apy: 17.127
  }
]
```
there is this `id` field which is unique. but it has these weird identifiers whcih i don't like. i suggest adding this at the back
of the contract, lowercased and without the empty space. that should be fine i think
```
{
    "type": "jar",
    "id": "pJar A",
...
}
```
otherwise lgtm!

PR #8, opened 2022-05-10
Harvest

waynebruce0x:


slasher125:
a few things:
- pls use `harvest-finance` instead of `harvest` as project folder and project name (otherwise we won't get the symbol on the UI)
- tvl as numbers instead of string
- same for apy (some seem to be strings)
- there is again 1 duplicated pool value, seems like we need to add an id again at the end
- pls map bsc -> binance

PR #9, opened 2022-05-10
add solend adaptor

0xodia:


src/adaptors/solend/index.js
```diff
@@ -0,0 +1,88 @@
...
+const configEndpoint = `${baseUrl}/v1/config`;
+const reservesEndpoint = `${baseUrl}/v1/reserves`;
+
+const TOKEN_LIST_URL = "https://cdn.jsdelivr.net/gh/solana-labs/token-list@main/src/tokens/solana.tokenlist.json"
+
+let tokenList
+
+async function getTokenList() {
+  if (!tokenList)
+    tokenList = (await utils.getData(TOKEN_LIST_URL)).tokens
```
0xodia: what's the best way to cache this? For now, I'm just following how it's done in https://github.com/DefiLlama/DefiLlama-Adapters/blob/c4e8a7c903bad4f2f6a6fb13e95c17d153866e3d/projects/helper/solana.js

src/adaptors/solend/index.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const buildPool = async (reserveConfig, reserveData) => {
+  const liquidity = reserveData.reserve.liquidity;
+  const coingeckoId =  (await getCoingeckoId(liquidity.mintPubkey)) ?? assetToCoinGeckoIdMap[reserveConfig.asset] ?? reserveConfig.asset;
+  const apy = Number(reserveData.rates.supplyInterest) + reserveData.rewards.reduce((acc, reward) => reward.side === 'supply' ? Number(reward.apy) + acc : acc, 0)
+
+  const newObj = {
+    pool: reserveConfig.address,
+    secondaryString: reserveConfig.marketName.charAt(0).toUpperCase() + reserveConfig.marketName.slice(1) + ' Pool',
```
0xodia: Ideally we can advertise our yields for an asset in all pools since some of our more attractive yields are in isolated pools.

image

Is it possible for "Project" column to say "Solend ([secondaryString])" so for example, "Solend (Stable Pool)" to differentiate USDC in different pools.

0xodia: Alternatively, we can simplify this by only showing our main pool assets for now if that's easier for you guys

0xngmi:
@slasher125

src/adaptors/solend/index.js
slasher125: Hello 0xodia. cheers for the adaptor! 

I suggest we keep all pools, not just main. 
Instead of adding the pool name to the project field, i suggest to add it to the symbol field instead like
```
symbol: `${utils.formatSymbol(coingeckoId)} (${secondaryString})`
```

Also, I would not use coingecko Ids, but instead the symbols from `reserveConfig.asset` so we get `BTC` instead of `bitcoin` which also means we could remove the tokenList related code

src/adaptors/solend/index.js
slasher125: the below suggestion would mean we can get rid of everything related to tokenList :)

slasher125:
2 things i noticed:
a) TVL's seem off when comparing against your UI, not sure why that is, but here is an example:

Screenshot 2022-05-11 at 18 17 47

compared to the adaptor output:

```
  {
    "pool": "BgxfHJDzm44T7XG68MYKx7YisTjZu73tVovyZSjJMpmw",
    "secondaryString": "Main Pool",
    "chain": "Solana",
    "project": "solend",
    "symbol": "usd-coin",
    "tvlUsd": 71352170.146748,
    "apy": 4.312874328706462
  },
```

0xngmi:
> TVL's seem off when comparing against your UI, not sure why that is, but here is an example:

I think its because total supply includes borrowed coins, thats pretty common in lending platforms

slasher125:
> 

yeah, makes sense, taking that one back :)

src/adaptors/solend/index.js
0xodia: i see. i was worried about downstream code being dependent on using coingeckoId for consistency and other code dependencies. For
[DefiLlama-Adapters](https://github.com/DefiLlama/DefiLlama-Adapters), that seemed important to get the correct price.

0xodia: ahh i see that the filtering seems to just a substring search which makes sense. I updated the PRs to ur suggestion

PR #12, opened 2022-05-11
Alpaca

waynebruce0x:


slasher125:
a few things on this on:

- lets change adaptor and project name to `alpaca-finance`
- both tvl and apy as numbers
- for farming pools: the apy shown is using 3x leverage, imo we should display the lowest of them all instead (1xleverage) which u can do via: `(farmRewardApr + tradingFeeApr) / leverage`
- for farming pools, can u make a change re symbol: instead of using utils.formatSymbol for farmingPools i think its better if we keep it like it is in `p.sourceName` but add the underlying project at the end in brackets, something: `“sourceName": "MDEX ETH-USDT" -> ETH-USDT (MDEX)`
- seems like there are 2 dupes as well on the pool for staking pools:

1b-alpaca-staking

think you could use the `stakingToken.address` from the object
```
        "fairLaunchStakingPools": [
            {
                "key": "ib-alpaca",
                "sourceName": "ALPACA Interest Bearing",
                "apr": "0.0",
                "apy": "0.0",
                "symbol": "ibALPACA",
                "tvl": "1589347.577801765482530221",
                "stakingToken": {
                    "address": "0xf1bE8ecC990cBcb90e166b71E368299f0116d421",
                    "symbol": "ibALPACA"
                }
            },
```
maybe we need to add the key at the end, but that should hopefully get rid of that duplicates after that step.

- for staking pools i would use `symbol` instead of `sourceName` as our symbol
```
        "fairLaunchStakingPools": [
            {
                "key": "ib-alpaca",
                "sourceName": "ALPACA Interest Bearing",
                "apr": "0.0",
                "apy": "0.0",
                "symbol": "ibALPACA",
```

src/adaptors/alpaca-finance/index.js
```diff
@@ -0,0 +1,71 @@
...
+        apy: Number(p.apy)
+    }));
+
+    const farmingPools = response.farmingPools.map(p => ({
+        pool: `${p.key}-farming-pool`,
+        chain: chain == 'bsc' ? 'Binance' : utils.formatChain(chain),
+        project: 'alpaca-finance',
+        symbol: formatSymbol(p.sourceName),
+        tvlUsd: Number(p.tvl),
+        apy: Number(p.totalApy / p.leverage)
```
slasher125: i think we need to use `(p.farmRewardApr + p.tradingFeeApr) / p.leverage` instead of `p.totalApy`.
at least when checking on the frontend it seems `totalApy` includes additional rewards which are not provided if leverage is 1x. So we can't scale it down like that.

using `(p.farmRewardApr + p.tradingFeeApr) / p.leverage` should then match exactly
whats being shown as `Total APR` like here (this is for the BNB-USDT pool on pancake swap):

Screenshot 2022-05-12 at 10 41 44
 
In this case we will show the correct 1x leverage APR, not APY. I'm not sure how they scale it up to APY in their case tbh, but i think its fine showing the APR here especially as its only slighly below APY.

src/adaptors/alpaca-finance/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+    const lendingPools = response.lendingPools.map(p => ({
+        pool: `${p.ibToken.address}-lending`,
+        chain: chain == 'bsc' ? 'Binance' : utils.formatChain(chain),
+        project: 'alpaca-finance',
+        symbol: utils.formatSymbol(p.symbol),
+        tvlUsd: Number(p.tvl),
+        apy: Number(p.totalApy)
+    }));
+
```
almost-llamas, line 60: It would be nice if you could take the Automated Vaults into account. The data can be accessed via `response.strategyPools`. 👍👍👍

slasher125: ty for that, we will those as well :)

waynebruce0x: > It would be nice if you could take the Automated Vaults into account. The data can be accessed via `response.strategyPools`. 👍👍👍

can you link to where these are on the UI? I couldnt find them, thanks

slasher125: Screenshot 2022-05-12 at 11 30 19

its up there, i couldn't find it either initially ^^

src/adaptors/alpaca-finance/index.js
slasher125: btw, to clarify, this is only relevant for the farming pools
i ran it locally, this seems to be good:
```
apy: (Number(p.farmRewardApr) + Number(p.tradingFeeApr)) / Number(p.leverage),
```

src/adaptors/alpaca-finance/index.js
```diff
@@ -0,0 +1,81 @@
...
+        apy: Number(p.apy)
+    }));
+
+    const farmingPools = response.farmingPools.map(p => ({
+        pool: `${p.key}-farming-pool`,
+        chain: chain == 'bsc' ? 'Binance' : utils.formatChain(chain),
+        project: 'alpaca-finance',
+        symbol: formatSymbol(p.sourceName),
+        tvlUsd: Number(p.tvl),
+        apy: (Number(p.farmRewardApr) + Number(p.tradingFeeApr)) / p.leverage
```
slasher125: coming back to this line for a last time. in order to scale this value to actual apy we will need to do the following:
```
apr = (Number(p.farmRewardApr) + Number(p.tradingFeeApr)) / p.leverage
compoundFrequency = 365 // for daily compounding
apy = ((1 + apr / 100 / compoundFrequency)**compoundFrequency - 1) * 100
```
makes sense?

if we do that trasnformation we end up getting exact apy values with leverage 1x:

Screenshot 2022-05-12 at 15 26 50

src/adaptors/alpaca-finance/index.js
```diff
@@ -37,10 +37,10 @@ async function apy(chain) {
         project: 'alpaca-finance',
         symbol: formatSymbol(p.sourceName),
         tvlUsd: Number(p.tvl),
-        apy: (Number(p.farmRewardApr) + Number(p.tradingFeeApr)) / p.leverage
+        apy: utils.aprToApy((Number(p.farmRewardApr) + Number(p.tradingFeeApr)) / p.leverage)
     }));
 
-    const ausdPools = response.ausdPools.map(p => ({
+    const ausdPools = response.alpaausdPools.map(p => ({
```
slasher125: not sure why u changed that key name

waynebruce0x: ffs mustve bum typed it

slasher125: lol

slasher125:
lgtm

ricardo-alpaca:
We were about to code this and this PR is here. Thank you sincerely from Alpaca Finance team.

PR #13, opened 2022-05-11
Vesper

waynebruce0x:


slasher125:
- seems to me we need to index "30" instead of "3" for the apy values
- `totalValue` gives native token units, we need to scale to usd via:
```(Number(p.totalValue) / Number(p.asset.decimals)) * p.asset.price```
- i took a look at their UI, i think we need to transform the symbol a bit in case its from an Earn pool, lets do this instead:
```
    symbol: v.name.startsWith('ve')
      ? `${v.name.split('-')[0]} (earn ${v.name.split('-')[1]})`
      : utils.formatSymbol(v.name),
```
this will make it easier to distinguish between the different symbols

slasher125:
lgtm

PR #15, opened 2022-05-12
vector

waynebruce0x:


slasher125:
Overall looks good, one thing which we'll need to address is:

Screenshot 2022-05-13 at 13 05 52

These isolated pools seem like 2-sided pools? 
this if from the adaptor output, taking `USDC-UST` as an example:
```
  {
    "pool": "vector-USDC_UST-staking",
    "chain": "Avalanche",
    "project": "vector-finance",
    "symbol": "USDC-UST",
    "tvlUsd": 32893478.240191445,
    "apy": 4.738178522681641
  },
```
Looks like we are just taking into account the first symbols liqudity without adding in the other. 
Also, it seems to me that the user can add either USDC or UST into that double sided pool, each given a different APY...
not sure how to deal with that tbh.

PR #16, opened 2022-05-12
minor fixes for pickle adaptor

MisterMard:
Fixes made summary:
- `pool` assigned `asset.details.apiKey` which is guaranteed to always be unique, slash-free and whitespace-free.
- `symbol` assigned with the more readable `asset.depositToken.components` which lists constituent tokens of the jar's depositToken.
- Include some missing jars (nested under `standaloneFarms`, `brineries` and `external` categories within the API response).
- Add extra conditions in `aggregateApys` to account for the missing jars structure.


slasher125:
Ty a lot that u took the time cross checking, appreciate it! We will get in those proposed changes. 

For the pool id change though we'll need to stick to the current version. Bit of background: the id needs to be unique not just within an adaptor but over all adaptors. the pool value is then only used as a reference for querying the corresponding historical data from the DB so we can show historical values. (this is improvable, potentially via hashing we are aware. will be improved!)

PR #17, opened 2022-05-16
angle

tutacrypto:
## Angle

All test passed. 

### Main index
Here is the main index for Angle boosted pools APR. 
As for Curve in your front, the max boost was used. 
Feel free to let it be knowed to users with the toggles on your front. (something like: _This is the APR with max boost from veANGLE locked 4 years._)

#### Adding Uniswap fees APRs
_- [x] adding uniswap pools fees APR_

Edit: Uniswap trading fees are included in the API already, and the other ones will be in a few days.

### DIFF
--- a/src/adaptors/angle/index.js
+++ b/src/adaptors/angle/index.js
@@ -6,18 +6,29 @@ const networks = {
 };
 
 let apyData;
+let symbol;
 const getPoolsData = async () => {
   const apyData = await utils.getData('https://api.angle.money//v1/incentives');
 
   const result = [];
   for (const staking of Object.keys(apyData)) {
-    if (apyData[staking].deprecated) continue;
     // the identifier is the voting gauge address and not the address of the staking pool
+    if (apyData[staking].deprecated) continue;
+    // changing the symbols so they fit the Defillama framework
+    symbol = apyData[staking]?.name.replace('/', '-').split(' ');
+    if (symbol.length == 1) {
+      symbol = symbol[0];
+    } else if (symbol.length == 2) {
+      symbol = symbol[0] + ' ' + symbol[1];
+    } else {
+      symbol = symbol[1];
+    }
+
     const pool = {
       pool: apyData[staking]?.address, // address of the staking pool
       chain: networks[apyData[staking]?.network] || 'Other',
       project: 'angle',
-      symbol: apyData[staking]?.name,
+      symbol: symbol,
       tvlUsd: apyData[staking]?.tvl,
       apy: apyData[staking]['apr']?.value,
     };
--- a/src/adaptors/angle/index.js
+++ b/src/adaptors/angle/index.js
@@ -1,24 +1,26 @@
+const { AppStream } = require('aws-sdk');
 const utils = require('../utils');
+
+const networks = {
+  1: 'Ethereum',
+  137: 'Polygon',
+};
 
 let apyData;
 const getPoolsData = async () => {
-  const apyData = await utils.getData(
-    'https://new-app-api-dot-angle-1.ew.r.appspot.com//v1/apr'
-  );
-  const tvlData = await utils.getData(
-    'https://new-app-api-dot-angle-1.ew.r.appspot.com//v1/incentives'
-  );
+  const apyData = await utils.getData('https://api.angle.money//v1/incentives');
 
   const result = [];
-  for (const gauge of Object.keys(apyData['gauges'])) {
-    const address = apyData['gauges'][gauge]?.address;
+  for (const staking of Object.keys(apyData)) {
+    if (apyData[staking].deprecated) continue;
+    // the identifier is the voting gauge address and not the address of the staking pool
     const pool = {
-      pool: address,
-      chain: 'Ethereum',
+      pool: apyData[staking]?.address, // address of the staking pool
+      chain: networks[apyData[staking]?.network] || 'Other',
       project: 'angle',
-      symbol: gauge,
-      tvlUSD: tvlData.filter((gauge) => gauge.address === address)[0]?.tvl,
-      apy: apyData['gauges'][gauge]?.details.max,
+      symbol: apyData[staking]?.name,
+      tvlUSD: apyData[staking]?.tvl,
+      apy: apyData[staking]['apr']?.value,
     };
     result.push(pool);
   }
@@ -30,7 +32,3 @@ module.exports = {
   timetravel: false,
   apy: getPoolsData,
 };
-
-getPoolsData().then((result) => {
-  console.log(result);
-});
--- a/src/adaptors/angle/index.js
+++ b/src/adaptors/angle/index.js
@@ -1,4 +1,3 @@
-const { AppStream } = require('aws-sdk');
 const utils = require('../utils');
 
 const networks = {
@@ -19,7 +18,7 @@ const getPoolsData = async () => {
       chain: networks[apyData[staking]?.network] || 'Other',
       project: 'angle',
       symbol: apyData[staking]?.name,
-      tvlUSD: apyData[staking]?.tvl,
+      tvlUsd: apyData[staking]?.tvl,
       apy: apyData[staking]['apr']?.value,
     };
     result.push(pool);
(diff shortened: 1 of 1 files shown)


tutacrypto:
PS: only the boosted pools are showing up now, but we are planning to upload all of them in the API in the near future.

slasher125:
Ty for the contribution!

-> i suggest we use min boost value instead (independet of the curve adaptor, i will change that one)
-> shouldn’t we also include the fees and interest values as on your UI? ideally we are displaying the sum of those:



```
    "sanFEI/EUR": {
        "value": 6.736723845801067,
        "details": {
            "min": 6.736723845801067,
            "max": 16.84180961450267,
            "fees": 1.5890000000000002,
            "interests": 1.3073819642938955
        },
        "address": "0x7c0fF11bfbFA3cC2134Ce62034329a4505408924"
    },
```
so min + fees + interests?

tutacrypto:
You're welcome!

Sure, makes sense. I used the max as it's what was used for Curve. 
If you have helpers/question marks in your UI it'd be great to show the boosted yields as well.

Don't know why I missed this, thanks for pointing it out! We're actually updating our API right now, so I'll do it sometime next week and ping you once it's done. 

Thanks for the feedback!

tutacrypto:
hey @slasher125 , just updated the PR taking into account your comments, and updated the endpoint. 
Let me know if it's fine with you :)

slasher125:
hello @tutacrypto. great! was just running it locally but the api gives me a 502, not sure why :/

src/adaptors/angle/index.js
```diff
@@ -0,0 +1,34 @@
+const { AppStream } = require('aws-sdk');
```
slasher125: think we can remove that import right?

src/adaptors/angle/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const result = [];
+  for (const staking of Object.keys(apyData)) {
+    if (apyData[staking].deprecated) continue;
+    // the identifier is the voting gauge address and not the address of the staking pool
+    const pool = {
+      pool: apyData[staking]?.address, // address of the staking pool
+      chain: networks[apyData[staking]?.network] || 'Other',
+      project: 'angle',
+      symbol: apyData[staking]?.name,
```
slasher125: re symbol:

I have noticed that the symbols are like `"Uni-V2 agEUR/FEI LP"`. One thing we do on the server side is adding additional fields to each pool such as if its a stablecoin, if its single exposure etc. 
The functions adding that info use the symbol as an input arg. 
If we rely on `apyData[staking]?.name` then the output of those additional funcs will be wrong.

We can do 2 things here:
a) we change the symbol at the adaptor level
b) we add logic on the backend

if we go with a) we'd need to make the following change:

```
  symbol = apyData[staking]?.name.replace('/', '-').split(' ');
  // eg "Uni-V2 agEUR/FEI LP" keep -> 'agEUR/FEI' or "FRAX/EUR Perp" -> 'FRAX/EUR'
  symbol = symbol.length > 2 ? symbol[1] : symbol[0];
```
I know its not ideal cause we loose some info of the specific pool names but should still be good informative enough for the user to identify the correct pool. let me know what u think

src/adaptors/angle/index.js
```diff
@@ -0,0 +1,34 @@
...
+      tvlUSD: apyData[staking]?.tvl,
```
slasher125: key should be `tvlUsd`

src/adaptors/angle/index.js
tutacrypto: Thanks for the feedback! I fixed the two other issues. 

About the symbols, I'm not sure what is the best solution here. Basically, we have three types of staking right now: 
- staking for LP tokens
- staking for sanTokens (deposit USDC/DAI/..., get sanDAI/EUR that can be staked) -> always single exposure
- rewards distributed to users opening Perpetual positions from one of the collateral -> exposed to USD/EUR price changes

I feel like doing a) would make it clearer for `#2` if it explains that it's a single exposure, but less so for `#1` and `#3` so probably not ideal. 

In any case, users will have to go through our frontend where we make an effort to explain each situation's context in great details, so you might not even need to add it yourself. (if you find anything unclear there please let me know btw)

Do you have an example of how this context looks like maybe?

slasher125: ty for the info on your side! One question: why is `sanDAI/EUR` token "single exposure"? if i deposit dai, shouldn't i also be exposed to agEUR in that case? 

--------------- 
Generally, let me try making things a big clearer regarding the toggles we have on the yield page:

Screenshot 2022-05-25 at 12 34 42

`Stablecoins`: A pool is exposed to stablecoins only eg. DAI on Aave, USDC-DAI on Uniswap etc.
`Single Exposure`: A pool is exposed to 1 token only. eg DAI on Aave, stETH on lido etc.
`No IL`: eg if pool consists of 1 token only or if more than 1 then they have to track the same underlying asset eg DAI-USDC-USDT on curve would be labelled with No IL (under the assumption of no depeg...given recent events I might actually change this, but thats at least what we do at the moment)

Lets map Angle's 3 different types of staking:
1) LP: `"Uni-V2 agEUR/FEI LP"` would be: `Stablecoin: True`, `Single Exposure: False`, `No IL: True` 
2) sanToken: `"sanFEI/EUR"`: would be: `Stablecoin: True`, `Single Exposure: False`, `No IL: True`
3) Perp: `"USDC/EUR Perp"` would be: `Stablecoin: True`, `Single Exposure: False`, `No IL: True`

Lmk if that makes sense and I got it right, especially for 2)

tutacrypto: You're welcome! 

When depositing DAI in Angle the sanDAI/EUR means that your DAI has been deposited in the pool over-collateralizing agEUR, as opposed to any other potential stablecoin we might launch. This DAI will be reinvested in Aave for example, but never pooled with agEUR. Thus, it is a single token exposure. 

I changed the two aspects that I think should be different: 
1. LP: `"Uni-V2 agEUR/FEI LP"` would be: `Stablecoin: True`, `Single Exposure: False`, `No IL: True` -> should be `False` (as agEUR and FEI are pegged to two different values, EUR and USD)
2. sanToken: `"sanFEI/EUR"` would be: `Stablecoin: True`, `Single Exposure: False` -> should be `True`, `No IL: True`
3. Perp: `"USDC/EUR Perp"` would be: `Stablecoin: True`, `Single Exposure: False`, `No IL: True`

On 3), Perp are not necessarily based on stablecoins. wETH/EUR Perp for example would be `Stablecoin: False`, as collateral used is wETH.

slasher125: got it. ty for explaining!

re the second section, yes that makes sense. regarding 3) thats ok, we will handle those cases so no worries there.

I think we are at the same page now. Which brings us back to how we could handle this on the adaptor side:

```
- remove substrings before/after empty space like Uni-v2, LP, Perp
- replace `/` with `-`
- and for symbols including `san` i would keep only the deposit token eg `sanFEI/EUR` -> `FEI`
```

if we change it like that then everything would integrate nicely with the backend.

Downside of this approach is that we loose a bit of info on the symbols/pool names. i know its not ideal, we are doing similar things for other projects as well. also, i'm actually considering to change this in the near future to have the full poolname as well. 
for now though, i don't think its a big problem if we apply the suggested steps, cause the user will check out your site anyways. But let me know if u disagree):

tutacrypto: Great! 

About doing the change, I think it's fine for 1) and 2), as you probably have other LP staking pools listed so users should understand. 

However, I think it would get really confusing for 3) as users could easily take these for liquidity pools or deposits whereas it is a very different mechanism. 

Maybe we can try to split the cases at the adaptor level to fix this for 1) and 2) and leave the full `Perp` names?
If splitting isn't too heavy code wise let's go for this :)

slasher125: Yeah thats fine by me :) I will handle the case for `Perp` on the backend

tutacrypto: Great, I just pushed the changes!
I did a simple `if ... else if`, let me know if there's a better way to do this and I can update the code :)

PR #20, opened 2022-05-19
Robovault

koffeebot:


slasher125:
Ty ser, couple of things:
- please change the folder name and project field value to `robo-vault` instead, thats the slug we use for your project.
- shouldn't we remove retired pools?

src/adaptors/robovault/index.js
```diff
@@ -0,0 +1,22 @@
+const utils = require('../utils');
+
+const poolsFunction = async () => {
+  const vaultDataFantom = await utils.getData('https://api.robo-vault.com/vaults/fantom');
```
SmoothBot: You can use `https://api.robo-vault.com/vaults` to get all the vaults across chains rather than calling for each chain

PR #21, opened 2022-05-20
Add CompliFi yield adaptor

MioGreen:
CompliFi protocol has launched V2 version and it is time to show the APY to the world :)


MioGreen:
Change API urls. The new PR is following.

PR #27, opened 2022-05-27
fix-LM-calculations

iagoLD:
I have changed the method for capturing TVL and changed the BPT price calculation for the aprLM function.

### DIFF
--- a/src/adaptors/balancer/index.js
+++ b/src/adaptors/balancer/index.js
@@ -50,6 +50,8 @@ const query = gql`
       id
       tokensList
       totalSwapFee
+      totalShares
+      totalLiquidity
       tokens {
         address
         balance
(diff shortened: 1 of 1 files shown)


slasher125 (commented):
Line 187, shouldn't this be changed as well using `x.totalShares` ?

slasher125:
Re prices:
Subgraph usd prices usually are calculated by using the main token (eg ETH), and deriving the USD price from the eg ETH-DAI pair.then that usd price is used for other pairs to calculate usd value. However, sometimes things get crazy on pairs with usually some tiny shitcoin like this one:

Screenshot 2022-05-28 at 09 46 55

Check the matic price of 1 disney on the top right. and it then uses that value and multiplies it with the 1trillion disney supply balance...leading to 290billion in tvl.

This is the reason why we use coingecko prices and not subgraph's usd values. Makes sense?
From a brief look at the output this doesn't seem to be an issue for balancer, but still its quite annoying if we get pools with super inflated tvl all of sudden on the frontend. (we are guarding against outliers, but still)

Aside from that i also noticed that when calculating tvl using coingeckos prices the tvl usd value's are closer to whats being shown on the balancer UI.

So, i suggest we revert the changes re tvl usd values and stick to the coingecko method for now :)

slasher125:
saw that u made the changes in a new PR, closing this one. ty again, was a great catch!

PR #36, opened 2022-06-07
feat: add ankr liquid staking yields

ittikorns:
Official Ankr Staking: https://www.ankr.com/staking/stake/


slasher125:
gm sir. ty for the contribution! I just checked in what we are doing for the protocol adaptor, seems like we only include the c-tokens. I'd like to keep it consistent in that sense and would prefer if we remove the reward earning b-tokens. otherwise PR looks great!

ittikorns:
@slasher125 bTokens are removed.

PR #38, opened 2022-06-09
Add RatAlert liquidity pools

george-prime:
Hi,

I have created this pull request to add RatAlert's newly created liquidity pools on Polygon. Would be grateful if you could merge it with your main repository to add our pools to Defillama pools list.

Thank you!
Best regards
george-prime
RatAlert team

### DIFF
--- a/src/adaptors/ratalert/index.js
+++ b/src/adaptors/ratalert/index.js
@@ -1,0 +1,46 @@
+const utils = require('../utils');
+
+const poolsFunction = async () => {
+
+  const ffoodData = await utils.getData(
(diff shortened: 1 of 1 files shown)


slasher125:
Hey, sorry for late reply. adaptor looks fine. however, on yields we are having a bit of a limitation in a sense that we currently only display pools with a tvl > 10k. while we can merge the PR now, at least with the current setup it won't show up on the frontend until the pool fulfills the above criteria.
this is not ideal, i am aware of that :/

there a few reasons for the above, but mainly its related to how frontend pages are built and certain limitations on data size.

PR #40, opened 2022-06-14
Change pool name and fetch new number from new data structure

molefund:
Dear DefiLlama, 
we have changed pool name and fetch new numbers from new data structure. Please help to review and help to merge. Thank you very much :)

### DIFF
--- a/src/adaptors/mole/index.js
+++ b/src/adaptors/mole/index.js
@@ -39,8 +39,8 @@ async function apy(chain) {
     apy: Number(p.apy) * 100,
   }));
 
-  const hedgePools = response.hedges.map((p) => ({
-    pool: `${p.key}-hedge-pool`,
+  const fundPools = response.funds.map((p) => ({
+    pool: `${p.key}-fund-pool`,
     chain: utils.formatChain(chainMapping[chain]),
     project: 'mole',
     symbol: utils.formatSymbol(p.iuToken.symbol),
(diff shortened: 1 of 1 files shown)


slasher125:
changing existing pool values introduces a bit of workload on our side cause I'd have to update the data base entries overwriting the current pool values with the new ones here. Otherwise the history on the charts would get truncated. lmk if really neceassary, if yes I'll make the changes

molefund:
Dear Slasher,

Thank you for your promote reply. Sorry to bring the inconvenience workload to you. In this change, we change not only the pool name, but also the new fetched data structure. The critical code as following: 

 >>>>  const fundPools = response.funds.map((p) => ({   

We get the data from "response.funds" . And the old data structure can't be fetched data now. 

Could you please help to update the data base entries for the new changed pool by any chance? I am sorry to bring you the inconvenience. I'd be very appreciate for your hard work :)

Best regards,
Mole 


> changing existing pool values introduces a bit of workload on our side cause I'd have to update the data base entries overwriting the current pool values with the new ones here. Otherwise the history on the charts would get truncated. lmk if really neceassary, if yes I'll make the changes

slasher125:
ah yeah i thought of just leaving the `pool: `${p.key}-fund-pool` as is but changing the rest...no worries though, i will make the change.

molefund:
> ah yeah i thought of just leaving the `pool: `${p.key}-fund-pool` as is but changing the rest...no worries though, i will make the change.

Dear Slasher,

That's great! It really nice. Thank you so much  O(∩_∩)O~

PR #45, opened 2022-06-22
Add pools config

jianfengxiaojianfeng:


jianfengxiaojianfeng:
> gm. please remove the `package.json` change, seems like that is not needed

Ok, done

PR #49, opened 2022-06-29
added Kokoa Finance yield server

jkkim111:
Kokoa Finance is the first cryptoasset-backed stablecoin platform built on the Klaytn network. Its stablecoin, KSD, is an entirely permissionless overcollateralized stablecoin. KSD strives to expand its use cases, including the “Earn Product” launched within Kokoa Finance. Users can deposit their KSD to start earning the Earn APY yields. In the backend, the deposited KSD are converted to dKSD whose balance is dynamically increasing proportional to the collateral managment yields. The APY is realized upon withdrawal of KSD.


slasher125:
gm, ty for adding! just checked your website and was wondering if you could include govern and farm pools as well? do they require any lockup period?

jkkim111:
Hey, yeah sure, we can add govern and farm pools. Govern pool requires 9
days to unlock, but none for the farm pools.
I will make another PR including all three of them then!

slasher125:
Nooice

PR #51, opened 2022-07-01
Added adapters: Venus.io and Scream.sh

glebskr:
Hi,
Added two new adapters:
- [venus.io](https://app.venus.io/)
- [scream.sh](https://scream.sh/)

### DIFF
--- a/src/adaptors/scream/index.js
+++ b/src/adaptors/scream/index.js
@@ -1,0 +1,37 @@
+const { request, gql } = require('graphql-request');
+
+const utils = require('../utils');
+
+const API_URL = 'https://api.thegraph.com/subgraphs/name/0xc30/scream';
+
+const query = gql`
+  {
+    markets {
+      id
+      underlyingSymbol
+      supplyRate
+      cash
+      underlyingPriceUSD
+    }
+  }
+`;
+
+const getApy = async () => {
+  const marketsData = await request(API_URL, query);
+
+  const pools = marketsData.markets.map((market) => ({
+    pool: market.id,
+    chain: utils.formatChain('fantom'),
+    project: 'scream',
+    symbol: market.underlyingSymbol,
+    tvlUsd: market.underlyingPriceUSD * market.cash,
+    apy: Number(market.supplyRate) * 100,
+  }));
+
+  return pools;
+};
+
+module.exports = {
+  timetravel: false,
+  apy: getApy,
+};
--- a/src/adaptors/venus/index.js
+++ b/src/adaptors/venus/index.js
@@ -1,0 +1,24 @@
+const utils = require('../utils');
+
+const API_URL = 'https://api.venus.io/api/governance/venus';
+
+const getApy = async () => {
+  const markets = await utils.getData(API_URL);
+  const marketsData = markets.data.markets;
+
+  const pools = marketsData.map((market) => ({
+    pool: market.address,
+    chain: utils.formatChain('binance'),
+    project: 'venus',
+    symbol: market.underlyingSymbol,
+    tvlUsd: market.totalSupplyUsd,
+    apy: Number(market.supplyApy) + Number(market.supplyVenusApy),
+  }));
+
+  return pools;
+};
+
+module.exports = {
+  timetravel: false,
+  apy: getApy,
+};


src/adaptors/pancakeswap/index.js
```diff
@@ -0,0 +1,189 @@
...
+const MASTERCHEF_ADDRESS = '0xa5f8C5Dbd5F286960b9d90548680aE5ebFf07652';
+
+const BSC_BLOCK_TIME = 3;
+const BLOCKS_PER_YEAR = (60 / BSC_BLOCK_TIME) * 60 * 24 * 365;
+
+const web3 = new Web3(RPC_URL);
+
+const pairQuery = gql`
+  {
+    pair(id: "<PLACEHOLDER>") {
```
0xngmi: why not use gql variables?

vrtnd: Fixed

src/adaptors/venus/index.js
```diff
@@ -0,0 +1,24 @@
...
+const getApy = async () => {
+  const markets = await utils.getData(API_URL);
+  const marketsData = markets.data.markets;
+
+  const pools = marketsData.map((market) => ({
+    pool: market.address,
+    chain: utils.formatChain('binance'),
+    project: 'venus',
+    symbol: market.underlyingSymbol,
+    tvlUsd: market.totalSupplyUsd,
```
slasher125: `tvlUsd` needs to be of type number.
my fault cause i accidentally pushed a change to `test.js` into master which actually would check for that.

Also. for venus, a lending protocol, lets use Liquidity instead as tvl (so supplyUSD - borrowUSD)

src/adaptors/scream/index.js
```diff
@@ -0,0 +1,37 @@
...
+const getApy = async () => {
+  const marketsData = await request(API_URL, query);
+
+  const pools = marketsData.markets.map((market) => ({
+    pool: market.id,
+    chain: utils.formatChain('fantom'),
+    project: 'scream',
+    symbol: market.underlyingSymbol,
+    tvlUsd: market.underlyingPriceUSD * market.cash,
+    apy: Number(market.supplyRate) * 100,
```
slasher125: seems like scream is offering rewardAPY. i suggest we add that here as well

src/adaptors/pancakeswap/index.js
```diff
@@ -0,0 +1,186 @@
...
+          bnbPrice,
+          cakePrice,
+          ethPrice,
+          pairInfo.token0.decimals,
+          pairInfo.token1.decimals
+        )
+          .div(1e18)
+          .toString();
+        const pool = {
+          pool: pairInfo.name,
```
slasher125: suggest we use something stronger than the poolname. i can see that you are pulling pool id values, that should be good?

src/adaptors/pancakeswap/index.js
```diff
@@ -0,0 +1,186 @@
...
+          chain: utils.formatChain('binance'),
+          project: 'pancakeswap',
+          symbol: pairInfo.name,
+          tvlUsd: reserveUSD,
```
slasher125: needs to be number dataype

src/adaptors/venus/index.js
vrtnd: Fixed

src/adaptors/pancakeswap/index.js
```diff
@@ -0,0 +1,186 @@
...
+  if (token0.includes('ETH')) return reserve0.times(ethPrice).times(2);
+  if (token1.includes('USD')) return reserve1.times(2);
+  if (token1.includes('BNB')) return reserve1.times(bnbPrice).times(2);
+  if (token1.includes('Cake')) return reserve1.times(cakePrice).times(2);
+  if (token1.includes('ETH')) return reserve1.times(ethPrice).times(2);
+};
+
+const getBaseTokensPrice = async () => {
+  const prices = await utils.getData(
+    'https://api.coingecko.com/api/v3/simple/price?ids=pancakeswap-token%2Cbinancecoin%2Cethereum&vs_currencies=usd'
```
slasher125, line 87: fyi: we have an internal price api https://defillama.com/docs/api mirrowing coingecko prices but without rate limits. no need to change this here, just something i wanted to point out and which u might be interested in the future

src/adaptors/pancakeswap/index.js
```diff
@@ -0,0 +1,186 @@
...
+        )
+          .div(1e18)
+          .toString();
+        const pool = {
+          pool: pairInfo.name,
+          chain: utils.formatChain('binance'),
+          project: 'pancakeswap',
+          symbol: pairInfo.name,
+          tvlUsd: reserveUSD,
+          apy: calculateApy(
```
slasher125, line 174: from a brief check on frontend, seems like there is some delta btw adaptor apy and UI eg:

```
  {
    "pool": "USDT-WBNB",
    "chain": "Binance",
    "project": "pancakeswap",
    "symbol": "USDT-WBNB",
    "tvlUsd": "167047154.66379022044044984363",
    "apy": 2.3555793949912465
  },
```

vs:
Screenshot 2022-07-05 at 11 59 21

src/adaptors/loopring/index.js
```diff
@@ -0,0 +1,23 @@
...
+const API_URL = 'https://api.loopring.network/api/v2/amm/poolsStats';
+
+const getApy = async () => {
+  const poolsData = await utils.getData(API_URL);
+
+  const pools = poolsData.data.map((pool) => ({
+    pool: `${pool.market}-pool`,
+    chain: utils.formatChain('ethereum'),
+    project: 'loopring',
+    symbol: pool.market,
```
slasher125: lets get rid of the AMM- prefix: so from `AMM-ETH-USDC` -> `ETH-USDC`

src/adaptors/loopring/index.js
```diff
@@ -0,0 +1,23 @@
+const utils = require('../utils');
+
+const API_URL = 'https://api.loopring.network/api/v2/amm/poolsStats';
+
+const getApy = async () => {
+  const poolsData = await utils.getData(API_URL);
+
+  const pools = poolsData.data.map((pool) => ({
+    pool: `${pool.market}-pool`,
```
slasher125: ideally we get something better than this as id if available

src/adaptors/pancakeswap/index.js
```diff
@@ -0,0 +1,186 @@
...
+            normalizedCakePerBlock,
+            cakePrice,
+            reserveUSD
+          ),
+        };
+        return pool;
+      })
+    )
+  );
+  return pools;
```
slasher125: there are 2 nulls in the final pools array, i suggest removing them here

src/adaptors/dforce/index.js
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  const pools = markets
+    .map(({ network, data: { supplyMarkets } }) => {
+      return supplyMarkets.map((market) => ({
+        pool: market.address,
+        chain: utils.formatChain(network),
+        project: 'dforce',
+        symbol: market.underlying_symbol,
+        tvlUsd: Number(market.supplyValue) / 1e18,
```
slasher125: would use liquidity here too instead of supply

src/adaptors/dforce/index.js
```diff
@@ -0,0 +1,44 @@
...
+        network,
+        data: await utils.getData(getApiUrl(NETWORKS[network])),
+      };
+    })
+  );
+
+  const pools = markets
+    .map(({ network, data: { supplyMarkets } }) => {
+      return supplyMarkets.map((market) => ({
+        pool: market.address,
```
slasher125: seems like this pool `0xb3ab7148cCCAf66686AD6C1bE24D83e58E6a504e` is repeated, occurs twice (on Abritrum and Polygon). We could add the chain id as suffix

src/adaptors/defichain-dex/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+const getApy = async () => {
+  const poolsData = await utils.getData(API_URL);
+
+  const pools = poolsData.data.map((pool) => ({
+    pool: `${pool.symbol}-pool`,
+    chain: utils.formatChain('defichain'),
+    project: 'defichain-dex',
+    symbol: pool.symbol,
+    tvlUsd: pool.totalLiquidity.usd || 0,
```
slasher125: needs to be number again

src/adaptors/defichain-dex/index.js
```diff
@@ -0,0 +1,23 @@
+const utils = require('../utils');
+
+const API_URL = 'https://ocean.defichain.com/v0/mainnet/poolpairs?size=200';
+
+const getApy = async () => {
+  const poolsData = await utils.getData(API_URL);
+
+  const pools = poolsData.data.map((pool) => ({
+    pool: `${pool.symbol}-pool`,
```
slasher125: if we don't have addresses for these, I'd rather add the project as suffix than `-pool`

PR #53, opened 2022-07-01
Adding friktion yield adapter

0xconvexity:
adapter for friktion vaults
https://friktion.fi/

### DIFF
--- a/src/adaptors/friktion/index.js
+++ b/src/adaptors/friktion/index.js
@@ -1,0 +1,49 @@
+const axios = require("axios");
+const utils = require('../utils');
+
+async function tvl() {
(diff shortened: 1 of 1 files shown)


slasher125:
ty for the contribution!
One thing I'd change is the symbol content in ():
Instead of `"BTC (mainnet_income_call_btc)"` I think `"BTC (Covered Call)"` is much easier to read and understand. Also, makes it easier for the user to check out the details on friktion and find that pool.

So some mapping from raw `poolId` to:
{
"mainnet_income_call_eth": "Covered Call",
"mainnet_income_put_sol": "Chash secured put",
...
}
would be nice.

PR #55, opened 2022-07-02
feat: Add atlendis.io adapter

smol-ninja:


smol-ninja:
Fixed TVL value to only include assets locked in the protocol.

PR #61, opened 2022-07-05
Add capsa pool

t0mm4rx:
Hello,
I would like to add the capsa.finance yield-bearing token to the DefiLlama Yield dashboard.

Here is the implemented and tested connector.


slasher125:
henlo t0mm4rx. ty for adding!
We are mostly using 24h periods when calculating fees/rewards which we scale up to a year. In your case, looking at the api response, I'd prefer if we use `last_week_performance` and scale that to a year instead of `projected_apy` which takes into account the last 3 months.

t0mm4rx:
Thank you for your quick feedback.
Ok, so I'll change to last_week_performance times 52.14.

slasher125:
great! * 52 is good enough :)

t0mm4rx:
Just tested and committed "last_week_performance" * 52 instead of "projected_apy".

slasher125:
can you pls remove the yarn.lock from the PR?

t0mm4rx:
Sure, done. Could be good to add it into the .gitignore btw.

slasher125:
> 

good call. will add that!

PR #109, opened 2022-07-20
Curve fix crnch

crnch:
Rebased on master; replaces #103 @glebskr @slasher125


slasher125:
niiiice! ty sir

crnch:
sweet data coming in 🤝😎

PR #135, opened 2022-07-26
feat: add hop-protocol yields

0xdapper:


0xdapper:
Need RPCs set for the tests to run, because tvl is computed by checking on-chain balance

glebskr:
Hi, thx for your pr.
It probably fails because u are passing `'gnosis'` to defillama/sdk as a chain try using `'xdai'` instead

SRC: https://github.com/DefiLlama/defillama-sdk/blob/master/src/general.ts#L41

0xdapper:
thanks for the suggestion @glebskr, seems to be provider error now

src/adaptors/hop-protocol/index.js
```diff
@@ -0,0 +1,115 @@
...
+    ].map((promise) => promise.then((data) => data.output))
+  );
+
+  return { tokenBalance, hopTokenBalance, decimals, tokenAddress };
+};
+
+const getPoolTvl = async (coreConfig, chain, token) => {
+  const { tokenBalance, hopTokenBalance, decimals, tokenAddress } =
+    await getPoolTokenBalances(coreConfig, chain, token);
+  const key = `${chain}:${tokenAddress}`;
```
slasher125: the adaptor test currently fails because the price is undefined for gnosis chain. you need to correct chain name, I'd just do:
```
  const key =
    chain === 'gnosis' ? `xdai:${tokenAddress}` : `${chain}:${tokenAddress}`;
```

src/adaptors/hop-protocol/index.js
```diff
@@ -0,0 +1,115 @@
...
+  const promises = tokens
+    .map((token) => {
+      const tokenPools = data[token];
+      const chains = Object.keys(data[token]);
+      return chains.map(async (chain) => {
+        const tvlUsd = await getPoolTvl(coreConfig, chain, token);
+        const tokenAddress = coreConfig[token][chain].l2CanonicalToken;
+        const hopTokenAddress = coreConfig[token][chain].l2HopBridgeToken;
+
+        return {
```
slasher125, line 97: and here, lets change the chain name to xdai too

```
        chainName = chain === 'gnosis' ? 'xdai' : chain;

        return {
          pool: `${chainName}-${token}`,
          chain: chainName,
...
```

src/adaptors/hop-protocol/index.js
```diff
@@ -0,0 +1,115 @@
...
+          pool: `${chain}-${token}`,
+          chain,
+          project: 'hop-protocol',
+          symbol: token,
+          apy: tokenPools[chain].apr * 100,
```
slasher125: lets use `apyBase` instead, seems like this is currently coming from trading feels only

0xdapper:
Why is it trying localhost for RPC?

0xdapper:
any idea @slasher125 ?

slasher125:
hey 0xdapper :) its been a very busy week for us. we haven't forgotten the PR though ;) going to take a look the updated PR asap

src/adaptors/hop-protocol/index.js
```diff
@@ -0,0 +1,118 @@
...
+          rewardTokens: [],
+          underlyingTokens: [tokenAddress, hopTokenAddress],
+          tvlUsd,
+        };
+      });
+    })
+    .flat();
+
+  const pools = await Promise.all(promises);
+  return pools.filter((pool) => !!pool.apy);
```
slasher125: the filter returns an empty array because you filter on `apy` instead of `apyBase` which breaks the tests. if u remove it it works fine. no need to remove pools with 0 apy

0xdapper: lol i didn't do any further debugging locally after seeing the localhost connection error which seemed very unrelated

0xdapper: seems like the error still persists, so are we just skipping CI for now and just merging this if works locally?

slasher125: don't know whats causing it, likely like rpc issue, but the xdai rpc we use should work fine. i'm going to merge this now. if it fails, i will ping you again :)

PR #144, opened 2022-07-28
Fix beefy slug

glebskr:


slasher125:
sry missed the pr here, made the change. closing

PR #312, opened 2022-09-08
add old field to response

slasher125:


slasher125:
return the pool field from the controller and store in the final s3 output so we can query this field is /poolsOld (next step)

PR #367, opened 2022-10-06
[morpho] implement 🦋  morpho-aave 👻 

julien-devatom:
Hi,
because morpho-aave is live for more than one month now, I'm implementing it to the yield server. 
Some points to mention:
- I moved the previous morpho integration (named as morpho into the code) to the morpho-compound folder and I have created a new folder for morpho-aave since the two protocols are independent. 
-  The problem of negative TVL for a given market is still here for morpho-compound, that's why I have done the same thing as morpho-compound: tvl = total supply. 
- I have also fixed an issue in morpho-compound integration (on total borrow)


julien-devatom:
I cannot check why the PR is falling (IMO is due to the renaming of the morpho folder)

slasher125:
gm. ty for updating!
tests are failing because of a check if morpho-compound and morpho-aave are valid/included in our protocols list. they aren't.
- Which means: you will need to keep the top level `morpho` folder with an index.js file as is. but feel free to either combine both compound and aave code into the same index file or keep them separate and import in `src/adaptors/morpho/index.js`
- to distinguish btw the 2 different markets use `poolMeta: Compound`, `poolMeta: Aave` instead

PR #464, opened 2022-11-30
Add Tortuga protocol on Aptos.

musically-ut:
[Tortuga protocol](https://tortuga.finance) is a liquid staking protocol built on Aptos.

The implementation is adapted from the source code from: https://docs.tortuga.finance/product-docs/protocol/reference

Only on-chain data is used, except the price of APT (Aptos chain's native token), which is read from Binance.

### DIFF
--- a/src/adaptors/tortuga/index.js
+++ b/src/adaptors/tortuga/index.js
@@ -6,11 +6,15 @@ const TGOV_ADDRESS =
   '0x84d7aeef42d38a5ffc3ccef853e1b82e4958659d16a7de736a29c55fbbeb0114';
 const NODE_URL = 'https://fullnode.mainnet.aptoslabs.com/v1';
 const RESOURCE_URL = `${NODE_URL}/accounts/${TGOV_ADDRESS}/resources`;
+const COINS_LLAMA_PRICE_URL = 'https://coins.llama.fi/prices/current/';
+const DECIMALS = 1e8;
+
+const aptosCoinName = 'coingecko:aptos';
 
 async function main() {
   const [resources, aptPrice] = await Promise.all([
     utils.getData(RESOURCE_URL),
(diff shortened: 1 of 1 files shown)


slasher125:
hi. ty for adding. tests keep failing. is the api being blocked when running form a server? eg it works fine when calling locally but doesn't here. the actual error: `reason: Unavailable For Legal Reasons`

We run all adapters on aws, so api's need to be callable server side.

musically-ut:
Ah, Binance API may not be available from inside the US maybe?

Let me see if I can get the APT price in USDC on chain too.

slasher125:
for prices u can just use our own api. one sec, will send u the snippet

llamatester:
The tortuga adapter exports pools: 
        
 
    Tests:       9 passed, 9 total
    Snapshots:   0 total
    Time:        0.311 s
    Ran all test suites.
    
    Nb of pools: 1
     
    
    Sample pools:
    ┌─────────┬──────────────────────────────────────────────────────────────────────────────┬─────────┬───────────┬────────┬───────────────────┬───────────────────┐
    │ (index) │                                     pool                                     │  chain  │  project  │ symbol │      tvlUsd       │        apy        │
    ├─────────┼──────────────────────────────────────────────────────────────────────────────┼─────────┼───────────┼────────┼───────────────────┼───────────────────┤
    │    0    │ '0x84d7aeef42d38a5ffc3ccef853e1b82e4958659d16a7de736a29c55fbbeb0114-tortuga' │ 'Aptos' │ 'tortuga' │ 'TAPT' │ 3691584.560995113 │ 6.361123857082298 │
    └─────────┴──────────────────────────────────────────────────────────────────────────────┴─────────┴───────────┴────────┴───────────────────┴───────────────────┘

PR #475, opened 2022-12-05
adaptor for polytrade yield :tada:

rohitpoly:


slasher125:
hi and ty for adding. a few things:
- please make use of on chain calls/subgraphs cause we are moving away from centralised api (see data sources in readme)
- are u guys listed on the main tvl dashboard yet?
- its generally advised to use apyBase and or apyReward instead of the high level apy field (we calculate the sum server side) so a user knows where the apy is coming from

karthik-polytrade:
@slasher125 Thanks for suggestions.

- Will go with this for now and will change to direct calls later. 
- We are live on mainnet, But we are not listed in defillama. 
- We have different calculation for the apy. So, that's the reason we gave totalApy and totalTvl.

slasher125:
you need to be listed on the tvl page first before listing on yields

rohitpoly:
@slasher125 how do I do that? Couldn't find anything to guide me.

slasher125:
https://github.com/DefiLlama/DefiLlama-Adapters#defillama-adapters

PR #510, opened 2022-12-22
Yield server for polytrade

rohitpoly:


src/adaptors/polytrade/index.js
```diff
@@ -0,0 +1,81 @@
...
+const { ethers } = require('ethers');
+const axios = require('axios');
+
+const getTokenPrice = async (token, exchangeTo, amount) => {
+ const { data } = await axios.get(`https://min-api.cryptocompare.com/data/price?fsym=${token}&tsyms=${exchangeTo}`);
+ const unitPrice = Object.values(data)[0];
+ return amount * Number(unitPrice);
+}
+
+const formatUnits = (amount, decimals) => Number(ethers.utils.formatUnits(amount, decimals));
```
karthik-polytrade, line 11: Add comments

src/adaptors/polytrade/index.js
```diff
@@ -0,0 +1,81 @@
...
+    chain,
+  });
+
+  const tokenPrice = await getTokenPrice(
+    "TRADE",
+    "USD",
+    formatUnits(tradeReward, '3'),
+  );
+
+const totalAPY = formatUnits(stableReward, '2') + tokenPrice;
```
karthik-polytrade, line 56: you can explain as a comment why we are adding.

src/adaptors/polytrade/index.js
```diff
@@ -0,0 +1,86 @@
...
+  const TRADE_REWARD = '0x64f33da516bf8289cf2f607aa357285753d6f039';
+  const STABLE_REWARD = '0x352A424Caf2aB698570b1E9a273209b5A0fF52BD';
+  const chain = 'polygon';
+  const abis = {
+    strategy: {"inputs":[],"name":"strategy","outputs":[{"internalType":"contract IStrategy","name":"","type":"address"}],"stateMutability":"view","type":"function"},
+    getReward: {"inputs":[],"name":"getReward","outputs":[{"internalType":"uint16","name":"","type":"uint16"}],"stateMutability":"view","type":"function"},
+    getBalance: {"inputs":[],"name":"getBalance","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},
+  };
+
+  const strategy = await sdk.api2.abi.call({
```
slasher125: not sure where u have seen`api2`, that doesn't exist. needs to be `api`

slasher125: strategy here returns `{ output: '0xfE624A12b1732d19680A7a2a2efBe21f1C0F3F19' }`

u need to call `.output` on sdk calls

src/adaptors/polytrade/index.js
```diff
@@ -0,0 +1,86 @@
...
+    target: LENDER_POOL_CONTRACT,
+    abi: abis.strategy,
+    chain,
+  });
+
+  const tvl = await sdk.api2.abi.call({
+    target: strategy,
```
slasher125: eg, here this will break cause needs to be strategy.output

src/adaptors/polytrade/index.js
```diff
@@ -0,0 +1,86 @@
...
+    abi: abis.getBalance,
+    chain
+  })
+
+  // Total liquidity in pool
+  const totalTVL = formatUnits(tvl, '6');
+
+  // Get rewards in TRADE token
+  const tradeReward = await sdk.api2.abi.call({
```
slasher125: same here, u need to call output (and on all other sdk calls too).

src/adaptors/polytrade/index.js
rohitpoly: @slasher125 I was saw `api2` in adaptor pr raise for TVL dashboard listing. I also saw the `.api.` and identified .output to the distinguishing factor

PR #519, opened 2023-01-03
Update BarnBridge contracts, subgraph and add Arbitrum

albertoramires:
* Update to the new versions of both the BarnBridge contracts and subgraphs.
* Added the Arbitrum chain.


albertoramires:
@slasher125 Sorry to bother you, but any idea why this is failing? Doesn't look like it's related to the changes I made in the PR, tests all pass locally, but I could be wrong. 🤔

PR #535, opened 2023-01-13
Add FiRM protocol

webmass:
Related to:
https://github.com/DefiLlama/DefiLlama-Adapters/pull/5045
https://github.com/DefiLlama/defillama-server/pull/1905

Adds FiRM's markets

I added a new attribute regarding lending: `isFixedBorrowRate`, hopefully we can then use this later on in the frontend to filter by "fixed-rate only", in the borrow aggregator for example, let me know what you think, the attribute name can be different ofc.

Code is using `queryFilter` which requires a provider supporting past events like alchemy, I didn't see a way to customize the provider in the env files, for my local tests I used a custom `AlchemyProvider` instead of the`providers.ethereum` you can see in the PR.

### DIFF
--- a/src/adaptors/inverse-finance-firm/abi.ts
+++ b/src/adaptors/inverse-finance-firm/abi.ts
@@ -1,0 +1,79 @@
+module.exports = {
+  "balance": {
+    "inputs": [
+    ],
+    "name": "balance",
+    "outputs": [
+      {
+        "internalType": "uint256",
+        "name": "",
+        "type": "uint256"
+      }
+    ],
+    "stateMutability": "view",
+    "type": "function"
+  },
+  "totalDebt": {
+    "inputs": [
+    ],
+    "name": "totalDebt",
+    "outputs": [
+      {
+        "internalType": "uint256",
+        "name": "",
+        "type": "uint256"
+      }
+    ],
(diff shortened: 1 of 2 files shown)


slasher125:
hi sir. 
- going to use a key of ours regarding the provider. so don't worry about that.
- a few smol things i'd change given u are a cdp:
```
1. add `mintedCoin` field and set it to "DOLA"
2. in case of cdp, tvlUsd = totalSupply
3. u can remove the reward related fields, apyReward, apyRewardBorrow and rewardToken
4. for cdp's, we require a debtCeilingUsd field indicating the maximum DOLA which can be minted (this is what we'll use as available liquidity on the UI)
5. lets remove the suggested field (cause it won't work on the server given its not defined in the db schema etc.); we gonna start with the meta tag you have already provided
```
lmk if unclear :)

webmass:
Hi,
Thanks for the review!
Sounds good I'll make the changes.

For item 5) I'll remove it for now but let me know if I can contribute (server/frontend repos etc) to make it happen, we'd really like to have this field and the ability to filter because otherwise it's very hard to appear in results in the borrow aggregator.

webmass:
feedbacks from review applied

PR #617, opened 2023-02-27
fix:rename bifrost liquid staking

killbond007:


slasher125:
henlo, pls revert the change on `pool` values, these shouldn't be updated cause it requires some additional work on the db side. ty

killbond007:
> henlo, pls revert the change on `pool` values, these shouldn't be updated cause it requires some additional work on the db side. ty

ok，for biforst Protocol DefiLlama page: https://defillama.com/protocol/bifrost-liquid-staking

PR #671, opened 2023-03-29
Added Goledo

0xn1c0:
Added Goledo Protocol to DeFiLlama Yields Dashboard. Goledo is a lending /borrowing protocol that runs on Conflux eSpace.


slasher125:
no idea why the test didn't run but checked locally and adapter looks ok

PR #682, opened 2023-04-05
Update BarnBridge adapter

albertoramires:
Update BB adapter to use the current contracts and subgraphs.


albertoramires:
@slasher125 Sorry to bother you, but any idea why this is failing? Doesn't look like it's related to the changes I made in the PR, tests all pass locally, but I could be wrong. 🤔

slasher125:
tested locally, ran fine for me too

PR #694, opened 2023-04-10
unshETH

wagmi33:


wagmi33:
Automated checks are failing I think bc config.env is not setup in the github, works fine locally.  Please suggest best way to setup environment.

slasher125:
looking gud, ty for adding 🙏

PR #706, opened 2023-04-17
Mellow Protocol pools on Polygon added

DmitryGr:


slasher125:
@DmitryGr henlo, can u confirm that all the polygon pools have negative apy values?

PR #710, opened 2023-04-18
AURA adapter should account for wstETH rewards

brightiron:
This adapter currently doesn't account for wstETH rewards. There are currently 13 pools w/ underreporting APYs

0WSTETH-50LDO
sfrxETH-stETH-rETH-BPT
B-stETH-STABLE
cbETH-wstETH-BPT
wstETH-ACX
B-cbETH-wstETH-Stable
50wstETH-50bb-euler-USD
wstETH-rETH-sfrxETH-BPT
50wstETH-50stk-APE
50wstETH-50ACX
50wstETH-50LDO
OHM-wstETH
50wstETH-50bb-a-USD


slasher125:
tyty for adding. one thing though: lets add the steth yield onto baseApy instead so its consistent with curve etc.

PR #713, opened 2023-04-19
auragi add yield-server

OxBill:


slasher125:
hi sir, pls refactor to on chain calls/subgraph instead of using api. ty

PR #723, opened 2023-04-23
fix: replace paraspace with paraspace-lending

jfzhou5:


slasher125:
henlo, just pushed the update myself (without changing the pool fields) closing this

PR #726, opened 2023-04-24
Adds yield adapter for Ensuro (https://ensuro.co)

gnarvaja:
Adds the yield adapter to compute Ensuro APY of Ensuro Pools (eTokens).

## About Ensuro 

Ensuro is a decentralized capital provider for insurance risk. We use smart contracts to curate competitive insurance portfolios. We allow anyone to invest in insurance risk and reap its benefits. By opening up the insurance market to new players, we fuel innovation for established and upcoming insurance partners.

https://defillama.com/protocol/ensuro

## Methodology

The adapter computes the totalSupply variation in the last day (excluding net deposits), as a percentage of initial totalSupply as the daily APR. Then converts the APR to annualized APY.

### DIFF
--- a/src/adaptors/ensuro/index.js
+++ b/src/adaptors/ensuro/index.js
@@ -1,0 +1,68 @@
+const sdk = require('@defillama/sdk');
+const { getProvider } = require('@defillama/sdk/build/general');
+const ethers = require('ethers');
+const axios = require('axios');
+
+const addressBook = {
+  polygon: {
+    usdc: "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174",
+    eTokens: [
+      // eTokens
+      {name: "eToken Junior Koala", address: "0x8d2Ee82c4172B2138B06b8037d769cBfAf9C0274", symbol: "eUSDJrKo"},
+      {name: "eToken Senior", address: "0x55bAe6690d46EA94D7F05DF7c80A85E322421fB6", symbol: "eUSDSr"},
+      {name: "eToken Innov Zone", address: "0x1C48Accaf6f8106883AA1973A45F02525652DEfC", symbol: "eUSDJrIZ"},
+      // Commenting out these two pools, since they aren't open to the general public for now
+      // {name: "eToken Junior Koala BMA", address: "0xBC33c283A37d46ABA17BC5F8C27b27242688DeC6", symbol: "eUSDJrKoBMA"},
+      // {name: "eToken Senior BMA", address: "0xF383eF2D31E1d4a19B3e04ca2937DB6A8DA9f229", symbol: "eUSDSrBMA"},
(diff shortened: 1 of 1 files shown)


gnarvaja:
Yes, sure. I will explain here, tell me if the explanation needs to be added somewhere else. 

The liquidity pools are used to provide solvency capital for portfolios of policies. 

Every time a policy is sold, an amount of solvency, proportional to the policy's maximum payout and the collateralization level, is locked in the protocol. This solvency comes from three sources:
1. Pure Premium: this is a fraction of the premium paid by the policyholder that's the estimated average loss for the policy.
2. Junior eToken: this is liquidity provided by LPs, in the junior pool of the specific risk portfolio.
3. Senior eToken: this is liquidity provided by LPs in the senior pool connected to the risk portfolio. Senior pools are diversified into several risk portfolios.

More info about solvency at https://docs.ensuro.co/product-docs/smart-contracts/policy#solvency-breakdown

For the duration of the policies, each liquidity pool needs to have enough liquidity to cover the solvency committed to active policies. The proportion of committed capital of the pool is represented by the *Utilization Rate*. **If Utilization Rate is under 100%, the LPs can withdraw their funds immediately without any lockup**.

When the Utilization Rate is at 100%, the LPs have to wait until the active policies expire to be able to withdraw their funds. But since when UR is at 100% the APY increases, this would potentially attract new LPs that will take UR under 100% allowing other LPs to exit the pool without waiting for the policy expiration.

You can check the *Unlock Schedule* metric on the eTokens pages: https://app.ensuro.co/eTokens/0x8d2Ee82c4172B2138B06b8037d769cBfAf9C0274

Regarding the risks, the LP funds are exposed to **unexpected** losses of the risk portfolios they are backing. In the normal evolution of the business, policy payouts are covered from the pure premiums collected. When losses are *less* than expected, the surplus is kept in the protocol as a contingency for future losses. When the losses are *more* than expected, the protocol first uses this surplus and only after that starts *borrowing* from the Junior eToken first and later (only if Junior is exhausted)  from the Senior eToken. This will be reflected as a negative return for the LPs.

Later when the losses are again under the expected losses (represented by collected pure premiums), the surplus is used to repay the loans taken with Senior eToken (first) and Junior eToken, giving an increased return on those pools. 

Check https://docs.ensuro.co/product-docs/smart-contracts/liquidity-pools#internal-loan for more information.

You can check on our website the past performance of the risk modules associated with the liquidity pools. You will see the performance is stable, the portfolio is well diversified in a large portfolio of policies, so unexpected losses have a very low chance to happen. For LPs wanting to take a conservative approach, is recommended to provide liquidity in the **Senior eToken** (https://app.ensuro.co/eTokens/0x55bAe6690d46EA94D7F05DF7c80A85E322421fB6).

PR #739, opened 2023-05-03
add Auragi yield

OxBill:


slasher125:
wrong slug sir, needs to be `auragi-finance`

PR #789, opened 2023-05-31
migrate chart to llamao

slasher125:


src/utils/lambda.js
```diff
@@ -11,26 +11,39 @@ function lambdaResponse(body, { statusCode = 200, headers } = {}) {
...
-      "Access-Control-Allow-Origin": "*",
-      "Cache-Control": `max-age=${cacheTime}`
+      'Access-Control-Allow-Origin': '*',
+      'Cache-Control': `max-age=${cacheTime}`,
     },
   };
 }
 
+const customHeader = (cacheTime) => {
+  return (
```
0xngmi: wont this just return 'header'?

you should return an array and then `...array` on `set()`

PR #887, opened 2023-07-20
Adding DFX Finance yield

ArieJones:
- Adding in adapter for DFX-Finance 
- Adding in goldsky subgraph 

Note: Ethereum LPs are based off of a veDFX boosted gauge model but we used a standard calculation without boosting involved in order to stick to showing a more accurate APY value for rewards for standard scenarios. All other chain pools are standard StakingRewardsMulti contracts.  

Let us know if we missed anything.


slasher125:
henlo, ty for adding, only thing i don't like is that for some pools eg this one: 
```  {
    pool: '0x38f818fcd57f8a1782bbcc1c90cb0fd03e7f0bd1',
    chain: 'Ethereum',
    project: 'dfx-finance',
    symbol: 'TRYB-USDC',
    tvlUsd: 121062.25340819421,
    rewardTokens: [ '0x888888435fde8e7d4c54cab67f206e4199454c60' ],
    apyReward: 22.562483591453518,
    apyBase: 0,
    apyBase7d: 0.9490688486129165,
    underlyingTokens: [
      '0x2c537e5624e4af88a7ae4060c022609376c8d0eb',
      '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'
    ],
    volumeUsd1d: 0,
    volumeUsd7d: 44190.928251000005
  },
```

apyReward is almost double compared to whats being shown on ui:
dfx

ArieJones:
Thanks!
Yeah this is because they are pools that are boosted. I am going to look at updating the calculation so that they fall at the bottom end of the range here.

ArieJones:
@slasher125 Ok I got some further clarification on these Ethereum pools. There is a special disbursement that is ongoing until mid Oct of 4090 DFX per gauge. So this is the discrepancy you are seeing in the low end APY on the site and the calculated APY. 

If you like I can always write some code around that to remove that value from the calculation.. but it is as of right now pulling the correct token amounts from the gauge and calculating out what a yearly APY would be correctly it seems. 

Thanks,
AJ

slasher125:
gm, as long as this special disbursement doesn't require any lockups/boosts, then its fine to include.

ArieJones:
> gm, as long as this special disbursement doesn't require any lockups/boosts, then its fine to include.

@slasher125 ok now I understand. I have updated the calculation on the gauge based pools to give an unboosted rewards apy. So it should be good to go now. 

Thanks for the insight!

PR #889, opened 2023-07-21
adds Gnosis to the Balancer V2 Yields Server

daveai:
adds yields on Balancer on Gnosis to the Balancer adapter / yield server


adamazad (approved):
lgtm

PR #898, opened 2023-07-28
Adds Vendor Finance v1 & v2 adapters

ForrestChew:
Both v1 & v2 yield adapters are very similar, with the primary difference being how to obtain the available lend balance in a pool. In v2, a function must be called on the pool itself to obtain the balance, whereas on v1, the balance can be obtained by calling the standard `balanceOf(pool.address)` function on the lend token contract.


slasher125:
hey, ty for adding. can u ping me on our discord re the env variable

ForrestChew:
> hey, ty for adding. can u ping me on our discord re the env variable

DM'ed

slasher125:
gm, just ran this. i've looked at the output, a few notes specific to vendor finance:
the way we build the array of objects for lending protocols is that:
- symbol = the collateral token
- apyBase = the supply apy for that token
- apyBaseBorrow = the borrow apy for that token
- ltv = the collaterals borrow power

in your case, it seems that the current objects are combinations of collateral and specific borrow token. rn i'm not sure what apyBase stands for. Is it the supply apy or the borrow apy? What is the term rate?

PR #900, opened 2023-07-28
✨ exactly: sum up rewards rates

itofarina:
The `RewardsController` is multi assets, each with a different rate. This PR sums up both rates.

### DIFF
--- a/src/adaptors/exactly/index.js
+++ b/src/adaptors/exactly/index.js
@@ -127,8 +127,7 @@ const apy = async () =>
                   chain,
                 });
                 /** @type number */
-                const rewardUsd = rewardsPrices[reward.toLowerCase()];
-
+                const rewardUSD = rewardsPrices[reward.toLowerCase()];
                 const firstMaturity = configStart - (configStart % INTERVAL) + INTERVAL;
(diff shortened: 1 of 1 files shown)


0xngmi:
this just removes the ceiling? could you elaborate?

itofarina:
> this just removes the ceiling? could you elaborate?

hi @0xngmi, sorry for the empty description. I've updated it.
In the above loop of the updated code, we get the rate per reward asset. The ceiling was my bad in the current implementation, and it became clear now that we are distributing 2 different assets. So the proper behavior is to sum up both rates as both rewards are being distributed in the same market. 

Let me know if wasn't clear. There's also more info in the [docs](https://docs.exact.ly/guides/protocol/rewardscontroller).

itofarina:
> this just removes the ceiling? could you elaborate?

hi @0xngmi, sorry for the empty description. I've updated it.
We get the rate per reward asset in the above loop of the updated code. The ceiling was my bad in the current implementation, and it became clear now that we are distributing 2 different assets. So the proper behavior is to sum up both rates as both rewards are distributed in the same market.

Let me know if it's clear. There's also more info in the [docs](https://docs.exact.ly/guides/protocol/rewardscontroller).

PR #913, opened 2023-08-04
feat: add arbitrove

deboguer-jng:


slasher125:
him ty for adding
- tvl should be fetched from chain as well

deboguer-jng:
> him ty for adding
> 
> * tvl should be fetched from chain as well

@slasher125 fixed.

slasher125:
apy needs to be scaled to %. also on your ui u show 77%, here the scaled apr is 34%. can u explain the difference?

deboguer-jng:
@slasher125 fixed the issue. the value from our dapp is hardcoded value.

PR #954, opened 2023-08-17
fix: Aura's missing pools on Optimism, Arbitrum, Polygon, Gnosis

Wertusser:
This PR updates Aura adapter to be able to fetch yield offers from networks other than Ethereum. It uses Aura's Subgraph  to get all enabled yield offers with APYs and TVL

### DIFF
--- a/src/adaptors/aura/index.js
+++ b/src/adaptors/aura/index.js
@@ -1,287 +1,162 @@
 const sdk = require('@defillama/sdk');
 const utils = require('../utils');
-const {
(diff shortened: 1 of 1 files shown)


slasher125:
ty, i'm not a fan of moving an adapter from on chain calls to subgraphs. would be great if u could refactor to extend it by still using the sdk :)

WardenJakx:
Hi @Wertusser I made an update in #971 so you are able to create a config for the chains you want and all pools will populate

PR #960, opened 2023-08-22
Sommelier: Remove subgraph dependency + Add TurboSWETH cellar

sandbochs:
# Ready for merge

## Description
This PR removes dependence on our subgraphs for calculating APY. We now get data directly from the RPC with the following flow:
- Determine block height for our daily and 7d APY intervals.
- Fetch share price @ those specified block heights.
- Use share price to calculate a daily and 7d APY.

Previously, we used hour snapshots of the share price when calculating daily APY in order to smooth out volatility. This method goes back to end of day snapshots so those APYs may be slightly different. We have double checked the 7d APYs and those are the same up to 3 decimal places. We feel this is a fair trade off since getting hourly snapshots would require 46 more RPC calls per Cellar.

### DIFF
--- a/src/adaptors/sommelier/config.js
+++ b/src/adaptors/sommelier/config.js
@@ -38,7 +38,7 @@ const stakingPools = {
   [realYieldBtc]: '0x1eff374fd9aa7266504144da861fff9bbd31828e',
   [defiStars]: '0x0349b3c56adb9e39b5d75fc1df52eee313dd80d1',
   [fraximal]: '0x290a42e913083edf5aefb241f8a12b306c19f8f9',
-  [turbosweth]: '0x8930ad2b661a3192d4b88f8bd02870a1db035589',
+  [turbosweth]: '0x69374d81fdc42add0fe1dc655705e40b51b6681b',
 };
 
 // List of v0815 Cellars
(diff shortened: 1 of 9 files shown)


src/adaptors/sommelier/config.js
```diff
@@ -261,6 +263,23 @@ const v2Pools = [
...
+    chain,
+    project,
+    symbol: 'WETH-SWETH',
+    poolMeta: 'TurboSWETH',
+    tvlUsd: 0,
+    apyBase: 0,
+    apyReward: 0,
+    rewardTokens,
+    underlyingTokens: [],
+    // FIXME: Double check url on launch
```
philipjames44: this is good to go

src/adaptors/sommelier/config.js
```diff
@@ -37,6 +38,7 @@ const stakingPools = {
   [realYieldBtc]: '0x1eff374fd9aa7266504144da861fff9bbd31828e',
   [defiStars]: '0x0349b3c56adb9e39b5d75fc1df52eee313dd80d1',
   [fraximal]: '0x290a42e913083edf5aefb241f8a12b306c19f8f9',
+  [turbosweth]: '0x69374d81fdc42add0fe1dc655705e40b51b6681b',
```
sandbochs, line 41: @philipjames44 verify the staker address

philipjames44: Verified.

philipjames44 (approved):

sandbochs:
Going to close this in favor of merging https://github.com/DefiLlama/yield-server/pull/961 first. Will reopen removing the subgraph dependency later since Turbo SWETH is launch tomorrow.

sandbochs:
Going to close this in favor of merging https://github.com/DefiLlama/yield-server/pull/961 first. Will reopen removing the subgraph dependency later since Turbo SWETH is launch tomorrow.

PR #962, opened 2023-08-24
Smardex: Add base network

estebgonza:
This pull request introduces Base on the `yiel-server` Smardex adapter.


estebgonza:
The tests are failing cause of the missing api key (that is already setup by your team).

PR #999, opened 2023-09-15
Fix/perpetual protocol adapter get prices

opass:
This PR fixed the pool adapter of perpetual protocol.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205494429945624


opass:
The test fails because this url `https://yields.llama.fi/distinctID` cannot fetch correctly in jest hook. 

Could defiLlama teams help to look at this issue?

https://github.com/DefiLlama/yield-server/blob/d4a9b0d64a5bd13586fd96421a968f826bc71a3c/src/adaptors/beforeTests.js#L40C58-L40C58

PR #1003, opened 2023-09-18
Sommelier: Add ETH Trend Growth vault

philipjames44:


philipjames44:
Errors in test dont look related, seem like global RPC setup steps failing

PR #1035, opened 2023-10-12
Sommelier: Add Turbo stETH vault

philipjames44:


philipjames44:
Errors in test seem like global RPC setup steps failing

slasher125:
hi, params are hardcoded to 0?

philipjames44:
> hi, params are hardcoded to 0?

The initial starting values are, same as all of the other vaults in this adaptor.

philipjames44:
Just for additional context, if necessary, all of the final data is populated from on chain data via RPC in the index.js file https://github.com/DefiLlama/yield-server/blob/master/src/adaptors/sommelier/index.js

PR #1111, opened 2023-12-11
Added Demeter V1 Adapter

bayou020:


slasher125:
henlo, ty for adding. there is a bug, `module.apy is not a function`, looks like u aren't exporting the correct func

slasher125:
closing for now, feel free to reopen with fix

PR #1140, opened 2024-01-12
Add Dinero

0xhafa:
Add apxEth apy


slasher125:
change first line to `const sdk = require('@defillama/sdk4');` and try again pls

PR #1143, opened 2024-01-15
Add rewardAPY

0xIvan:


slasher125:
hi, can u provide some details about the apy rewards? what is required to receive it? any lockups or other requirements?we only list lower bound apy values

0xIvan:
> hi, can u provide some details about the apy rewards? what is required to receive it? any lockups or other requirements?we only list lower bound apy values

It's single staking rewards that we currently give out because of the Arbitrum STIP program. No lockups.

PR #1164, opened 2024-01-31
SandClock: update APY calculation to align with sandclock app

fyang1024:
to align with app.sandclock.org


slasher125:
hi, on our end we use a 24h window for all our projects, so we should actually fix this to 1day instead of increasing it to 7day. lets make that change instead. also pls, keep the current 7day value and set it the `apyBase7d`

PR #1183, opened 2024-02-14
IPOR Protocol - Add wstETH on Arbitrum

amydlarz:


amydlarz:
If I set up the ETHEREUM_RPC environment variable, then the tests pass. If not, they fail in the same way

PR #1189, opened 2024-02-16
add blueprint

ruvlol:


slasher125:
hi, ty for adding.
why are u commenting out the sdk code blocks?

ruvlol:
The commented part was cloned from existing Merkl integration. Basically our thing is Merkl copy but with only Blueprint pools tracked.

slasher125:
k, if the sdk call works, then lets use that instead of the api

slasher125:
@ruvlol gm, are u still working on this?

slasher125:
closing this, feel free to reopen using the sdk instead

PR #1190, opened 2024-02-16
feat(adapters): added paladin-warlord apy adapter

0xMemoryGrinder:
Hello,

I made an adapter to integrate the Warlord product of Paladin Finance. Let me know if there is anything that needs to be changed.

Thank you in advance for your review !


0xMemoryGrinder:
I see that the CI is not passing due to as reverting transaction. I managed to solve this issue by using an alchemy RPC. Let me know if I should change nayhting to make it more compatible.

PR #1209, opened 2024-03-05
check token prices

daiwanwei:


slasher125:
sir the issue that we seem to have is that we can't call your api inside an aws lambda

PR #1239, opened 2024-03-20
Add Polygon network for yldr.com

kulikov-yldr:


slasher125:
did the tests work locally?

kulikov-yldr:
yes, now locally tests are pasing but didn't run them when opened that tbh, could it be the case that CI failed because https://github.com/DefiLlama/DefiLlama-Adapters/pull/9459 was not merged initially?

PR #1242, opened 2024-03-21
add permissionless vault data

daiwanwei:


slasher125:
gm, have u tested the adapter locally?

PR #1254, opened 2024-04-01
update: add more chains

eriscrypt:


slasher125:
pls change the import to `const sdk = require('@defillama/sdk5');`

eriscrypt:
all tests pass on my local machine

slasher125:
how can the tests pass if it fails for inEVM?
i need to update the sdk, for now, pls comment out the inEVM parts. i'll update this later

eriscrypt:
Screenshot 2024-04-03 at 13 17 45

slasher125:
i just ran this locally, its failing. if u check what it says above the output you should see the specific error
u got duplicated pool values

eriscrypt:
we have same addresses on each network ...

slasher125:
as mentioned in the readme, the pool values need to be unique, you can just add a chain suffix

PR #1386, opened 2024-06-18
Fix(umami-finance): Add back incentives APR for vaults

Clonescody:
Added back the logic to calculate ARB incentives APR in its own file

Misc : 
- Rename helper to match other files
- Removed $UMAMI related constants as the token isn't tracked for yield anymore (deprecated)
- Removed subgraph and API URLs as they're not in use anymore
- Clean up imports

URL to check datas : https://umami.finance/vaults


Clonescody:
Not sure what's the test failing is about 🤔 works on my computer
image

PR #1414, opened 2024-07-04
Add Yield for TLX

chase-manning:


llamatester:
The tlx-finance adapter exports pools: 
        
 
    Tests:       18 passed, 18 total
    Snapshots:   0 total
    Time:        0.279 s
    Ran all test suites.
    
    Nb of pools: 2
     
    
    Sample pools:
    ┌─────────┬──────────────┬────────────┬───────────────┬─────────────┬────────────────────┬─────────┬────────────────────┬──────────────────────────────────────────────────┬──────────────────────────────────────────────────┬──────────────────────┐
    │ (index) │     pool     │   chain    │    project    │   symbol    │       tvlUsd       │ apyBase │     apyReward      │                   rewardTokens                   │                 underlyingTokens                 │       poolMeta       │
    ├─────────┼──────────────┼────────────┼───────────────┼─────────────┼────────────────────┼─────────┼────────────────────┼──────────────────────────────────────────────────┼──────────────────────────────────────────────────┼──────────────────────┤
    │    0    │ 'tlx-staker' │ 'optimism' │ 'tlx-finance' │   'stTLX'   │ 1616989.7972513984 │    0    │ 124.03262596655287 │ [ '0x8c6f28f2F1A3C87F0f938b96d27520d9751ec8d9' ] │ [ '0xd9cc3d70e730503e7f28c1b407389198c4b75fa2' ] │     'TLX Staker'     │
    │    1    │ 'tlx-locker' │ 'optimism' │ 'tlx-finance' │ 'lockedTLX' │ 845579.9703427576  │    0    │ 100.59668064121605 │ [ '0xd9cc3d70e730503e7f28c1b407389198c4b75fa2' ] │ [ '0xd9cc3d70e730503e7f28c1b407389198c4b75fa2' ] │ 'TLX Genesis Locker' │
    └─────────┴──────────────┴────────────┴───────────────┴─────────────┴────────────────────┴─────────┴────────────────────┴──────────────────────────────────────────────────┴──────────────────────────────────────────────────┴──────────────────────┘

src/adaptors/tlx-finance/index.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const LOCKER_ADDRESS = '0xc068c3261522c97ff719dc97c98c63a1356fef0f';
+const TLX_ADDRESS = '0xd9cc3d70e730503e7f28c1b407389198c4b75fa2';
+const STAKER_ADDRESS = '0xc30877315f3b621a8f7bcda27819ec29429f3817';
+const SUSD_ADDRESS = '0x8c6f28f2F1A3C87F0f938b96d27520d9751ec8d9';
+
+const LOCKER_LAUNCH_DATE = new Date(1712731443000);
+const LOCK_DURATION = 26 * 7 * 24 * 60 * 60 * 1000; // 26 weeks
+const CHAIN = 'optimism';
+const SAMPLE_PERIOD_DAYS = 30;
```
slasher125, line 13: we calculate our yields over a 1day period, so, apyBase = 1d window, scaled to a year.
additionally, you can add apyBase7d

chase-mew: @slasher125  Fees are only charged periodically when rebalances happen. A 1 day window wouldn't really be useful for the user, as it would vary significantly day by day. As in, 0% one day, 1000% another day. I think it's okay to leave the sample period as 30 days, and have DeFiLlama calculate that daily. Essentially that is representing a 1 day calculation over our most accurate APR estimate (which happens to use 30 days).

This is also somewhat of an industry standard. Yearn for example does periodic harvests, and so has a similar situation. So [their adapter](https://github.com/DefiLlama/yield-server/blob/master/src/adaptors/yearn-finance/index.js#L15) returns the 30 day average (from the API), which DefiLlama then displays as the 1 day, as that is a snapshot of the most accurate APR Yearn can calculate on that day. If they returned the actual daily the yield chart would be a mess as they don't harvest every day, so the yield chat would be something like: 0%, 0%, 21%, 0%, 0%, 21%.

PR #1458, opened 2024-08-08
stellaswap v3 support

atticuzor:


slasher125:
@atticuzor gm sir, the adapter is sitting at root, pls move it inside /src/adaptors

PR #1501, opened 2024-09-09
drop lendos yields

northern05:
delete lendos yields


slasher125:
whats the reason here? we don't delete adapters

northern05:
Sorry, thats was a my mistake

ср, 11 вер. 2024 р., 09:10 користувач slasher125 ***@***.***>
пише:

PR #1541, opened 2024-10-09
Move risk rating to yield server

vrtnd:


src/api/controllers/risk.js
```diff
@@ -0,0 +1,43 @@
+const fetch = require('node-fetch');
+
+export const YIELD_RISK_API_EXPONENTIAL =
```
slasher125: export breaks, can also be removed given its not used anywhere else

slasher125:
we need to add `app.use(express.json());` right after https://github.com/DefiLlama/yield-server/blob/master/src/api/app.js#L17

src/api/controllers/risk.js
```diff
@@ -0,0 +1,43 @@
...
+  'https://api.exponential.fi/api/pool-risk/search';
+
+const getRiskRating = async (req, res) => {
+  const poolData = req.body;
```
slasher125: lets add a try catch in case their api is down etc. 

in catch i'd add something like; 
```
res.status(503).json({
  status: 'error',
  message: 'Exponential API unavailable',
});
```

PR #1576, opened 2024-10-29
Ionic-protocol Reward apy correction 

amish777:


amish777:
@slasher125 how much time will it take to display new values in lama ui as it is still showing old ones ?

PR #1612, opened 2024-11-18
Fixes chain from bnb to BCS

B-aublys:
For some reason there are conflicts... But we just changed the chain our pool is on


slasher125:
@B-aublys pushed here 1c57bcb1f8044296a7bbcc762bba2a35e8f19513

PR #1619, opened 2024-11-25
Change Uniswap v3 subgraph queries to use pagination

rav-ipor:
Uniswap v3 adapter is fetching only first 1000 pools from subgraph for every chain. Because of that it misses many pools with >10k TVL.
In this PR, adapter fetches all Uniswap v3 pools from subgraph with pagination and filter out pools with less then 10k TVL.
Uniswap v2 pools are filtered out, because some of them are duplicated in both subgraphs used in Uniswap v2 and Uniswap v3 adapters.


rav-ipor:
Tests are failing because of missing `GRAPH_API_KEY` env var:
```
ethereum ClientError: auth error: payment required for subsequent requests for this API key
```
I tested it on forked repo on copied `Test_Change_On_Demand` workflow from `Test_Change` workflow with added `GRAPH_API_KEY` env var and it passes tests, here is the link to the run:
https://github.com/IPOR-Labs/yield-server/actions/runs/12025729366/job/33523370299

slasher125:
@rav-ipor ty. issue with this version is runtime, which for me locally was almost 15min which is way to long. if u can explore a faster version of this that be appreciated

rav-ipor:
@slasher125 ok, I will try to speed it up

PR #1718, opened 2025-02-12
Smardex / USDN: Implements USDN yield calculation

estebgonza:
This PR implements the USDN yield in the SmarDex yield adapter.

- I registered two new ABIs: USDN Token and USDN Protocol.
- I implemented the USDN APR formula as calculated on the SmarDex website: [smardex.io](https://smardex.io/).


estebgonza:
As remember the `test.yml` workflow is not going through since the `SMARDEX_SUBGRAPH_API_KEY` is not well defined in the workflow env vars

PR #1728, opened 2025-02-17
Smardex / USDN: Fix USDN yield APR

estebgonza:
This pull request fixes USDN Yield APR. Percentage formatting issue, and decimal issue.


slasher125:
@estebgonza can confirm the apr value is in %?

estebgonza:
> @estebgonza can confirm the apr value is in %?

Exactly, the USDN APR value is in percentage.

**Current output on my PR:**

```ts
{
  pool: '0xde17a000BA631c5d7c2Bd9FB692EFeA52D90DEE2',
  symbol: 'USDN',
  project: 'smardex',
  chain: 'Ethereum',
  tvlUsd: 3147807.5164719927,
  apyBase: 0.44773911165071384
}
``` 

This corresponds to a 44.77% APR since the USDN protocol was released.

[c473118](https://github.com/DefiLlama/yield-server/pull/1728/commits/c4731183a1663dfa511d1b9e8af3d2cf0206b33d)

I made a change that led to a misunderstanding in my calculation. I was formatting the APR by multiplying by 100_000, which caused a decimal issue due to a mistake in my calculation. I wasn't dividing by the number of annual seconds, but by the number of days instead: it's fixed :)

Also, the tests are not successful because you need to use the `SMARDEX_SUBGRAPH_API_KEY` from your secrets in the env variable of the `tests.yml` workflow.

Thank you for your time @slasher125

slasher125:
> This corresponds to a 44.77% APR since the USDN protocol was released.
what i meant is to make sure to scale it to % in the adapter, u still need to multiply by 100

estebgonza:
> > This corresponds to a 44.77% APR since the USDN protocol was released.
> > what i meant is to make sure to scale it to % in the adapter, u still need to multiply by 100

Ok it's clear for the expected format, thanks for your answer

PR #1780, opened 2025-03-21
chore: update apy

janndriessen:
Updates APY fetching for Index products.

### DIFF
--- a/src/adaptors/index-coop/index.js
+++ b/src/adaptors/index-coop/index.js
@@ -21,7 +21,8 @@ const SetTokenABI = ['function totalSupply() external view returns (uint256)'];
 
 const buildPool = async (index) => {
   try {
-    const apy = await getApy(index.symbol);
+    const apy = await getApy(index.address);
+    console.log(apy);
     const tvlUsd = await getTvlUsd(index);
     const chain = utils.formatChain(index.chain);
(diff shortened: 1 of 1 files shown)


janndriessen:
gm @slasher125 please let me know if the failing test is due to the code change or on defillamma's side.

slasher125:
does it work locally?

slasher125:
seems like the bug is from `getSupply` func, can just use our sdk instead of ethers

janndriessen:
> seems like the bug is from `getSupply` func, can just use our sdk instead of ethers

thank you for the pointer good ser @slasher125. should be good for review now. 👍

src/adaptors/index-coop/index.js
```diff
@@ -37,14 +37,13 @@ const buildPool = async (index) => {
   }
 };
 
-const getApy = async (indexSymbol) => {
-  const indexPath = indexSymbol.toLowerCase();
+const getApy = async (address) => {
   const res = await superagent.get(
-    `https://api.indexcoop.com/${indexPath}/apy`
+    `https://api.indexcoop.com/v2/data/${address}?chainId=1&metrics=apy`
```
janndriessen, line 42: using our new api

PR #1782, opened 2025-03-21
Fix solidlyv3 pool tvl, add more chains

abanchev:


abanchev:
We have pools that share the same address, while being on different chains, i will append chain name to the pool property, but i am not sure if this is okay with your api

abanchev:
CI error likely caused by missing ALCHEMY_CONNECTION_ARBITRUM env variable

slasher125:
i've tried locally, but its breaking. can you confirm that the tests run fine on your end?

abanchev:
@slasher125 All is working here
image
in my env i have only GRAPH_API_KEY and ALCHEMY_CONNECTION_ETHEREUM

slasher125:
ah, its breaking here cause of the graph api key...cause alchemy one is set. anyways,
can u confirm that the pool values are the exact same addresses but just with the chain as prefix?

abanchev:
Yeah, we have couple of pools that end up having the same address across
chains because of multi chain tokens paired with the native wrapped that
also has the same address, the data looks correct

slasher125:
@abanchev 
sry had to revert this, its still breaking for me testing locally, even with keys set. 
generally, would be better if we don't rely on alchemy altogether and just use our sdk instead

abanchev:
@slasher125 i was just refactoring existing code that was already using alchemy
anyway i will rewrite it to use the sdk instead, it should not take long, and i will open another PR

PR #1828, opened 2025-04-22
fix(test): ignore casing when checking for duplicate pools

mastilver:


mastilver:
Some pools (24) are not unique when ignoring casing. Like those two:

436136113-9fad5bf9-edec-4bce-8054-36831eb93847

This PR fixes some https://github.com/DefiLlama/yield-server/pull/1827 by blacklisting Uniswap pools from Merkl

PR #1838, opened 2025-04-24
Blend sdk updates

Ryang-21:
- Update blend-sdk to v3.0.1
- Add new pools
- Use CoinGecko BLND price

### DIFF
--- a/src/adaptors/blend-pools/index.js
+++ b/src/adaptors/blend-pools/index.js
@@ -1,72 +1,80 @@
 const {
-  Pool,
+  PoolV1,
   BackstopToken,
   Backstop,
   FixedMath,
+  TokenMetadata,
 } = require('@blend-capital/blend-sdk');
-const { getPrices, keepFinite, formatChain } = require('../utils');
+const { getPrices, keepFinite, formatChain, getData } = require('../utils');
 
 const BACKSTOP_ID = 'CAO3AGAMZVRMHITL36EJ2VZQWKYRPWMQAPDQD5YEOF3GIF7T44U4JAL3';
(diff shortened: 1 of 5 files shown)


src/adaptors/blend-pools/index.js
```diff
@@ -1,72 +1,80 @@
...
       pools.push({
         pool: `${pool.id}-${reserve.assetId}-stellar`.toLowerCase(),
         chain: formatChain('stellar'),
         project: 'blend-pools',
-        symbol: reserve.tokenMetadata.symbol,
+        symbol: tokenMetadata.symbol,
         tvlUsd: totalSupply - totalBorrow,
-        // Supply is kept as APR to prevent overestimation of APY
-        apyBase: reserve.supplyApr * 100,
+        //Estimated weekly compounding
```
slasher125: if the pool doesn't auto-compound for the user, then lets keep using the previous fields

Ryang-21: The pool does auto compound but it requires the reserve/asset to be interacted with by some user ie. someone borrows or supplys which then triggers the reserve state to be updated.

Ryang-21: If thats still a concern I can switch to use the apr rate just let me know

slasher125: > If thats still a concern I can switch to use the apr rate just let me know

yeah better to use switch back to the non compounding fields, ty

Ryang-21: Sounds good that should be fixed. I left the borrow apy as the estimated apy with compounding as it should be better for users to have an overestimation on how much the borrowing costs.

PR #1845, opened 2025-05-01
fix merkl stuck pools

mastilver:


mastilver:
by not listing past pools and pools with a 0% reward, DL does not updates them

For example this pool
image
is stuck on 5.20% when it should be 0% as the campaign ended

Let me know if this is correct or if it's an issue with yield aggregator itself?

slasher125:
good catch, ty. 

generally we auto drop any pools form the API and UI if the last pool update is >7days old

PR #1853, opened 2025-05-06
fix Euler on Berachain

mastilver:


mastilver:
I would recommend reviewing without withespace: https://github.com/DefiLlama/yield-server/pull/1853/files?diff=split&w=1

I'm not sure what's happening here:
- it fails on local
- it fails on the CI
- it fails on https://berascan.com/address/0x36F611B6d77a1DfeeAFF8f144C24A91bb9CA2D5c#readContract#F6 (try inputing 0x4BbDA9A841d201A4D08d346355Fd9c1318351168)
- it looks like Euler pools are being updated on DefiLlama which shouldn't be possible:

0xngmi:
the issue in ci and local is that berachain free rpcs are broken, you can replace them by setting the env variable BERACHAIN_RPC

mastilver:
@0xngmi Thank you, I will close this PR then

PR #1855, opened 2025-05-06
added base, swell, unichain

leemount96:


0xngmi (approved):
ci failing is due to berachain rpcs

PR #1883, opened 2025-05-20
feat: update lista dao yield info

tyler-tsai:
Add new adaptor for protocol(https://defillama.com/protocol/lista-lending)

ref page: https://lista.org/lending#earn


tyler-tsai:
image

tyler-tsai:
> The 451 error occurs because the GitHub runner uses a US-based IP address, and our API doesn't allow access from the US. It works fine on your server since it's using a different IP.

@slasher125 could you please help have a look?

PR #1889, opened 2025-05-23
Merkl: remove past and uniswap pools

mastilver:
Uniswap is handled by https://github.com/DefiLlama/yield-server/pull/1827
I added past pools earlier in https://github.com/DefiLlama/yield-server/pull/1845


slasher125:
gm, past as in no longer active?

mastilver:
@slasher125 That's correct

I previously added them as otherwise they would be stuck at their previous apr until they disappear (now they are forever on 0%)
This PR removes them straight away, happy to remove the logic if you prefer

PR #1894, opened 2025-05-28
update basisos index

fullmetalholder:
This PR enables fetching the BasisOS APR (Rewars) for vaults.

We’re providing rewards in $BIOS tokens offchain, so in the code we’re using an API. More details can be found in the docs https://docs.basisos.org/tokenomics/liquidity-mining

This API is consistent across all vaults.


slasher125:
@fullmetalholder pls add missing rewardTokens field

PR #1909, opened 2025-06-09
XSY: adds APY tracking for yUTY token

brandonaaskov:
This adds APY tracking for the yUTY token vault contract.


slasher125:
gm @brandonaaskov 
the project slug doesnt match any of our existing project (see local test output), do u have a tvl adapter on our main tvl dashboard? if not, u need to create that first

PR #1952, opened 2025-06-30
Update index.js

i6opoga:
adding pools links


i6opoga:
also needs to change the directory to crosscurve instead of crosscurve-(by-eywa)

slasher125:
closing this (better to keep urls as they are rn)

PR #1959, opened 2025-07-01
yieldoor: add eth looped vault

d-grasshopper:
# Add Yieldoor's Ethereum Looped Vault

## Questions to reviewers

1. I've added a dependency to the Morpho SDK. Please advise if that's ok, if not, could you offer another solution (if you've perhaps been through a similar situation before)?
2. I've tried to optimize the many calls made, but any suggestions are welcome

### DIFF
--- a/src/adaptors/package-lock.json
+++ b/src/adaptors/package-lock.json
@@ -11,6 +11,8 @@
       "dependencies": {
         "@blend-capital/blend-sdk": "3.0.1",
         "@defillama/sdk": "^5.0.112",
+        "@morpho-org/blue-sdk": "^3.0.6",
+        "@morpho-org/blue-sdk-viem": "^3.0.0",
         "@stacks/network": "^6.13.0",
         "@stacks/transactions": "^6.15.0",
         "@ton/ton": "14.0.0",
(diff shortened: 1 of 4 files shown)


d-grasshopper:
Of course, we use their SDK, since they have all the necessary math and on-chain calls to calculate Morpho positions with interest up to a timestamp in their classes/helpers.

d-grasshopper:
Hi, yes, I've investigated it, I can reimplement most of the logic, but I will also have to add/implement some of their math lib, i.e - https://github.com/morpho-org/sdks/blob/59ba43e6a011a23291b9b2a4d75f5e9d730e38c1/packages/blue-sdk/src/math/AdaptiveCurveIrmLib.ts#L65.

So that's why I went with just importing their SDK. I'll leave the decision to you.

slasher125:
> Hi, yes, I've investigated it, I can reimplement most of the logic, but I will also have to add/implement some of their math lib, i.e - https://github.com/morpho-org/sdks/blob/59ba43e6a011a23291b9b2a4d75f5e9d730e38c1/packages/blue-sdk/src/math/AdaptiveCurveIrmLib.ts#L65.
> 
> So that's why I went with just importing their SDK. I'll leave the decision to you.

lets try an implementation without these dependencies, its fine if you copy over bits from their repo

slasher125:
gm, thx. we dont expose the graph key in our tests, i'll run ur code locally. can u confirm that it works for u using your own key?

PR #1972, opened 2025-07-04
Add ZKP2P Yield to DefiLlama

unhappyben:
We currently have all our data publicly available in Dune, so it was easier to use this than rebuild the wheel. I mentioned this in the discord and it seems okay. 

It caches the data so that if it is less than 24 hours old, we do not recall the API. I kept it to the latest day of data available, but if we need historical I can change the script. 

Our Methodology is : 

Our APR is calculated using a TVL-weighted median of pool-level APRs. Each pool represents a payment rail (e.g., Venmo, Zelle, Wise). Our APR calculatiuon can be found here https://docs.zkp2p.xyz/user-guides/for-sellers/calculating-apr. 

This is then rolled up and should give the Median APY as suggested by your docs. 

Let me know if you need more info.

### DIFF
--- a/src/adaptors/zkp2p/index.js
+++ b/src/adaptors/zkp2p/index.js
@@ -1,0 +1,76 @@
+const axios = require('axios');
+
+// Cache the data for 24 hours to avoid hitting Dune API limits
+let cachedData = null;
+let lastFetchTime = 0;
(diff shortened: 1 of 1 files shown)


unhappyben:
Going to reopen to pass test with dune api key

PR #1973, opened 2025-07-07
 Add ZKP2P Yield to DefiLlama

unhappyben:
**_Reopened with Dune API Key as variable to pass test_**

We currently have all our data publicly available in Dune, so it was easier to use this than rebuild the wheel. I mentioned this in the discord and it seems okay.

It caches the data so that if it is less than 24 hours old, we do not recall the API. I kept it to the latest day of data available, but if we need historical I can change the script.

Our Methodology is :

Our APR is calculated using a TVL-weighted median of pool-level APRs. Each pool represents a payment rail (e.g., Venmo, Zelle, Wise). Our APR calculatiuon can be found here https://docs.zkp2p.xyz/user-guides/for-sellers/calculating-apr.

This is then rolled up and should give the Median APY as suggested by your docs.

Let me know if you need more info.

### DIFF
--- a/src/adaptors/zkp2p/index.js
+++ b/src/adaptors/zkp2p/index.js
@@ -45,6 +45,7 @@ const getApy = async () => {
       tvlUsd: Number(latestData.tvl),
       apy: Number(latestData.protocol_apr_pct),
       url: 'https://zkp2p.xyz',
+      poolMeta: 'APR is a TVL-weighted median across payment rail pools (e.g., Venmo, Zelle). Details: https://docs.zkp2p.xyz/user-guides/for-sellers/calculating-apr',
     };
 
     console.log('ZKP2P: Fetched fresh data from Dune');


slasher125:
gm @unhappyben 
yes, test will fail here given the api key is not exposed during tests, i'll check locally.
as for the cache: we run yield backend on aws lambda, so the current cache solution won't work. u can remove all of that and i'll take care of it so we trigger this only N times a day

unhappyben:
gm @slasher125 - removed the caching! Thanks for your help

slasher125:
@unhappyben 
```
Nb of pools: 1
 

Sample pools: [
  {
    pool: 'zkp2p-main-pool-base',
    chain: 'Base',
    project: 'zkp2p',
    symbol: 'ZKP2P',
    tvlUsd: 94193.82060399998,
    apy: 25.3286,
    url: 'https://zkp2p.com'
  }
]```

wrong url? also, can u point me this pool on your ui, i can't find it

unhappyben:
Screenshot 2025-07-14 at 07 32 15
Screenshot 2025-07-14 at 07 32 30
Screenshot 2025-07-14 at 07 32 42


TVL comes from chain bc it is always most up to date, the TVL comes from Dune which is slightly behind - I just took these 3 screenshots, the difference is minimal

unhappyben:
GM @slasher125, just updated with poolMeta field! 

Thanks a lot for your help.

PR #1988, opened 2025-07-13
Revert name change Peapods Finance

NonFinancialAdvisor:
I'm not sure what's the reason for this name change as the protocol hasn't gone through a rebrand or anything and remains to be Peapods Finance.


Define101:
@NonFinancialAdvisor https://defillama.com/protocol/peapods-options
We changed the name

NonFinancialAdvisor:
@Define101 but why? The protocol didn't change name should still be Peapods Finance. I have no clue where this name 'options' came from and why it was modified.

slasher125:
done

PR #2007, opened 2025-07-21
list garden staking to yield

Yaswanth-Vempuluru-7916:


llamatester:
The garden adapter exports pools: 
        
 
    Tests:       9 passed, 9 total
    Snapshots:   0 total
    Time:        0.24 s
    Ran all test suites.
    
    Nb of pools: 1
     
    
    Sample pools:
    ┌─────────┬───────────────────────────────────────────────────┬────────────┬──────────┬────────┬─────────┬───────────────────┐
    │ (index) │ pool                                              │ chain      │ project  │ symbol │ tvlUsd  │ apy               │
    ├─────────┼───────────────────────────────────────────────────┼────────────┼──────────┼────────┼─────────┼───────────────────┤
    │ 0       │ '0xcbb7c0000ab88b473b1f5afd9ef808440eed33bf-base' │ 'Arbitrum' │ 'garden' │ 'SEED' │ 5489400 │ 17.27231108365837 │
    └─────────┴───────────────────────────────────────────────────┴────────────┴──────────┴────────┴─────────┴───────────────────┘

Yaswanth-Vempuluru-7916:
@slasher125 we have our apy calculation based on two contracts Staking and Distributor,  also the apy is dependent on our epoch wise fee and reward pool, so it is not possible for us to calculate using onchain calls,

slasher125:
> @slasher125 we have our apy calculation based on two contracts Staking and Distributor, also the apy is dependent on our epoch wise fee and reward pool, so it is not possible for us to calculate using onchain calls,

can u provide some additional details, why part of that can't be fetched from the contracts directly?

as for tvl that should def be fetched from on chain

slasher125:
> Is it possible to iterate over all logs of a smart contract to calculate the TVL for DefiLlama? As for APY, it cannot be determined on-chain.

yes, with sdk.api.util.getLogs [see examples in the repo]

> I noticed that other adapters are also fetching data directly from the API. Could you help me understand why we're being advised to use onchain in our case?

cause these are either older adapter or it was not possible fetching via sdk

we try to prevent the usage of apis where applicable cause they hide the methodology

PR #2026, opened 2025-07-31
feat(parallel-protocol-v3): add support to sUSDp yield token

FabienCoutant:
Add Parallel Protocol v3 USDp yield token (sUSDp)


slasher125:
@FabienCoutant can u pls share a link to ur ui where i can double check the values, also keep in mind we have a min tvl thr of $10k for the main dashboard

PR #2041, opened 2025-08-07
track Asymmetry Finance USDaf Stability Pool yields

pastelfork:
adapter is equivalent to Liquity's with the exception that WBTC branch collateral price is fetched using canonical WBTC address (actual collateral is a wrapper for decimal adjustment)


0xVertigo:
Gm @slasher125 We used the exact same naming convention as Liquity does for their SPs seen below

Screenshot 2025-08-07 at 10 51 35 AM

If it needs to be shortened still, please let me know as that's no problem, we'll get creative

PR #2115, opened 2025-09-23
fix: bug with aerodrome-v1 staked supply ratio

Pochetes:
## Overview
there's a possible error with Aerodrome V1's rewards APY calculation. the `stakedSupplyRatio` is actually inverted because `totalSupply` is the gauges supply of LP tokens and `poolSupply` is the aerodrome pool.

this could be an issue because if half of the LPs have staked, it actually doubles the yield. the discrepancy is less obvious as more LPs stake, but there is a visible overestimation of the APY for the newer pools. and since there's no point of being an unstaked LP, the bug disappears very quickly as the ratio nears 1.


0xkr3p:
thanks for the pr, but closing this as the existing code appears correct

PR #2141, opened 2025-10-08
feat(termmax): Refactor vault handling and add event interfaces for V1 and V2

tkspring-henry:


src/adaptors/termmax/index.js
```diff
@@ -57,297 +97,479 @@ async function getMerklOpportunities() {
...
-            .filter((a) => a);
-        }
-
-        pools.push(pool);
-      }
+    const vault = {
+      pool: `${address}-${chain.toLowerCase()}`,
+      chain,
+      project: 'termmax',
+      symbol: assetNames.output[i].output,
```
slasher125, line 490: we should keep the symbol, eg rn the adapter outputs: ` symbol: 'Wrapped liquid staked Ether 2.0',` should just be `WSTETH`

tkspring-henry: fixed on commit 903e57e

tkspring-henry:
@slasher125 I've fixed the issue you found. Please review it when you are available. Our partners would like to see the vault statistics, but the pull request has been held back for over a week, so it has become urgent now.

PR #2153, opened 2025-10-15
add enclabs yields (Sonic)

Numerikat:
add enclabs yields for Sonic chain


slasher125:
@Numerikat pls add apyBaseBorrow and ltv info so we can include your protocol in our borrow aggregator. ty

PR #2177, opened 2025-10-28
add zealous swap pools

louissaadgo:
Zealous Swap (Kasplex) — Pools adapter for DefiLlama yield-server
Pulls per-pool TVL (USD) and APR directly from our public API.

Endpoint used: https://kasplex.zealousswap.com/v1/pools
Notes:
 - We use `apr` from our API as `apyBase` (already annualized %).
 - If `apr` is missing, we fall back to fee APR from volume * feeRate.
 - We include `apyReward` only if our API reports a positive `farmApr`.
 - Only emit pools with `hasUSDValues === true` and tvl > 10000 to avoid noise.


slasher125:
hi,

> tvl > 10000 to avoid noise.

pls remove, we handle this server side

- poolMeta seems to be constant, in which case it doesn't really add any information. please remove
- pls fetch as much info as possible form on chain calls/subgraphs

PR #2194, opened 2025-11-10
Fix nest-credit adaptor: add APY fallback logic and required fields

kristinatub:
- Set apyBase to 7-day APY with fallback to 30-day APY if empty
- Default apyBase to 0 only if both 7-day and 30-day APY are empty
- Add required 'project' field set to 'nest-credit'
- Remove invalid field names (apy30day, apy7day)
- Add live=true parameter to API URL
- Handle empty/missing APY values gracefully

### DIFF
--- a/src/adaptors/nest-credit/index.js
+++ b/src/adaptors/nest-credit/index.js
@@ -1,0 +1,66 @@
+// adapters/nest-vaults.js
+const utils = require('../utils');
+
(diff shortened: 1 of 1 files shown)


slasher125:
> * Set apyBase to 7-day APY with fallback to 30-day APY if empty
> * Default apyBase to 0 only if both 7-day and 30-day APY are empty

pls do not mingle different windows, we can keep the 7day values but not the 30day otherwise it just gets confusing

PR #2200, opened 2025-11-13
adding kaspacom-dex yields

giladm11:
Adding kaspacom-dex yields.
The info is coming from our subgraph


giladm11:
> hi @giladm11 url doesn't work

hi, you are right, I had a type error. it's fixed now. sorry

PR #2228, opened 2025-12-08
getting yields info from kcom backend

giladm11:
project: kaspacom-dex
We need to change the way we retrieve the info since we are going to block access from the graph node in the future.


slasher125:
@giladm11 
apyBase for `'WKAS-BMT'` seems to be inflated, its ~1.6% in the current version from the graph

slasher125:
> we took only the last 6 days in the graph version, because it's only daily info and calculated by that.
in our api, we actually sum up all the sales in the requested timeframe, which give more accurate information. so it will add the sales of the first day by the hour.

and this leads to a difference btw 17% ? your ui is still tracking 0.2% hmm

PR #2236, opened 2025-12-17
add hermes-v2 adapter and implement gauge staking APR tracking

0xBugsy:
Quick question, should we provide a start blockNumber to enable timetravel? Thanks in advance.


0xBugsy:
@slasher125 should be fixed now, sorry for the wait

PR #2254, opened 2026-01-06
feat: add kerne protocol yield adapter

enerzy17:
## Release Notes

* **New Features**
  * Added Kerne Protocol Yield Adapter with live APY and TVL data fetching for the Base chain vault, displaying real-time yield metrics on kLP tokens from Kerne Finance.


slasher125:
@enerzy17 make sure we have a tvl adapter first, otherwise we can't list on yields. also pls check the test output

PR #2257, opened 2026-01-07
update shmon apy to 6 day window

marshabl:
Monad epochs are 5.5 hours, meaning some days there are 4 epochs and some 5. This impacts shmonad yield for a 24 hour period causing it to look more spiky or volatile than reality. Updating the window from 24 hours to 6 days smooths the spikes and gives a truer representation of the current yield.

### DIFF
--- a/src/adaptors/shmonad/index.js
+++ b/src/adaptors/shmonad/index.js
@@ -19,24 +19,24 @@ const SHMONAD_ABI = {
 };
 
 // APY window in days
-const APY_WINDOW_DAYS = 6;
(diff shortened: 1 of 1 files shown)


slasher125:
hi @marshabl looking our current adapter and the history, its seems quite stable though (excluding early dec): https://defillama.com/yields/pool/ee40513c-9356-4c53-9f26-446b484a8ae2

i suggest keep apyBase as is, but you can add apyBase7d using a 7day window (instead of 6). this is an optional field which can enabled via the dropdown on the main pool page

PR #2259, opened 2026-01-08
Fix Liminal adapter: use 7-day APY as apyBase

shkkgit:


slasher125:
why the update of apyBase = apyBase7d?

shkkgit:
> why the update of apyBase = apyBase7d?

NAV isn't updated on a daily basis, therefore computing the APY based on a 24 hour PPS increase isn't accurate

PR #2260, opened 2026-01-09
add mstable-v2 project

paveltimosh:


slasher125:
have made the db updates, should be good in an 1-2h on the ui

PR #2630, opened 2026-04-27
Merkl balancerv3

franzns:
show merkl rewards on pools in yield dash. We dont have the reward tokens available for merkl rewards in our API so I used ZERO address, hope that is fine.

## Summary by CodeRabbit

* **New Features**
  * Added support for additional reward token type in APR calculations.

* **Bug Fixes**
  * Improved handling of empty reward token lists with fallback value.


0xkr3p:
hi @franzns, apologies for the wait - thanks for the PR, we need to have legitimate reward tokens to move forwards with this. Are you able to use the merkl helper instead to match pool with their api?  There are two helpers which may be of use: `merkl-additional-reward.js` and `merkl-by-identifier.js`

franzns:
Thanks for the tip, I'll have a look!

PR #2633, opened 2026-04-27
feat: allow negative apy via per-adapter flag

### LINKED ISSUE #2409: Feature Request: Add support for negative APY
Some yield assets may have negative APY (losses) over some periods of time, currently it seems DefiLlama does not support this and report this.

kaankolcu:
closes #2409
adds an opt-in `allowNegativeApy` flag on adapter modules so loss-bearing pools (eg liquid restaking depegs, leveraged stable strategies) can report real negative apy instead of being clamped to 0. existing options-category and hardcoded-list behaviour is preserved, so no current adapter changes output.

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration flag enabling adapters to report and retain negative APY values when explicitly enabled
  * Expanded APY filtering range to include negative yields, allowing pools with negative returns to be recognized and displayed
  * Updated documentation for the new negative APY configuration option

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ module.exports = {
   timetravel: false,
   apy: apy, // Main function, returns pools
   url: 'https://example.com/pools', // Link to page with pools (Only required if you do not provide url's for each pool)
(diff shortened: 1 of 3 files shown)


slasher125:
hi, unfortunately i'll have to close this, we are working on a new server which will handle this case

PR #2653, opened 2026-05-06
 Update APY calculation to include landing sales revenue

trxadm:
Updated the APY calculation for TR-Energy. We have added landing_percent to the baseApy.

This parameter represents the additional yield generated from retail energy sales via our landing page. This revenue is distributed to users, so including it provides a more accurate representation of the total real-time APY for stakers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected APY calculation for the TRX pool to include the previously omitted landing component. APY values now incorporate profit, static, and landing portions with sensible defaults, so displayed yields are more accurate and may change for affected pools.

### DIFF
--- a/src/adaptors/tr-energy/index.js
+++ b/src/adaptors/tr-energy/index.js
@@ -22,12 +22,18 @@ async function apy() {
   const stats = await getStats();
   const trxPrice = await getTrxPrice();
 
+  const {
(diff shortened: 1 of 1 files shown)


0xkr3p (approved):
hi @trxadm, thanks for the PR - approved

PR #2705, opened 2026-05-26
Add BTCD sBTCD yield adapter

z-j-lin:
Adds the BTCD yield adapter for the Ethereum sBTCD staking vault only.

Exposes the sBTCD ERC-4626 vault
Computes TVL from on-chain totalAssets()
Computes base APY from 30-day previewRedeem() growth
Uses the BTCD underlying token and vault address metadata required by Yield Server

## Summary by CodeRabbit

* **New Features**
  * Added BTCD adapter providing TVL and a 30‑day APY estimate for the sBTCD staking vault and linking to https://btcd.fi (deep-link to stake page).
  * Derives an algorithmic BTC peg from on‑chain price feeds to normalize TVL and APY calculations.
  * Shows vesting/cooldown duration labels on the vault when available.
* **Robustness**
  * Skips pools or reports zero APY when price, TVL, or timing inputs are missing or invalid.

### DIFF
--- a/src/adaptors/btcd/index.js
+++ b/src/adaptors/btcd/index.js
@@ -185,14 +185,15 @@ const apy = async () => {
       `SBTCDOracle.latestRoundData().answer must be positive, got ${priceUsd_e8}`
     );
   }
-  // tvlUsd = totalSupply × price / 10^(18+8) = USD raw with 8 decimals.
+  // tvlUsd = totalSupply × price / 10^(18+8) = USD raw with 8 decimals
+  // (the oracle's `decimals()` is fixed at 8 per IPriceOracle).
   // Split the BigInt into whole+fractional dollars so we don't lose cents
   // of precision once total TVL grows past Number.MAX_SAFE_INTEGER (~$90M
   // when expressed at 8-decimal scale).
   const tvlUsdScaled = (totalSupply * priceUsd_e8) / 10n ** 18n; // 8-decimal USD
-  const HUNDRED_M = 10n ** 8n;
-  const wholeUsd = tvlUsdScaled / HUNDRED_M;
-  const fracUsd = tvlUsdScaled % HUNDRED_M;
+  const ONE_USD_E8 = 10n ** 8n; // = 1.00 USD expressed in 8-dec oracle units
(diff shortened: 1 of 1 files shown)


src/adaptors/btcd/index.js
```diff
@@ -0,0 +1,226 @@
...
+ * @returns {Promise<number>} APR as a percentage (e.g. 11.88 for 11.88%).
+ */
+async function computeApyBase() {
+  let latest, old;
+  try {
+    latest = await sdk.api.util.getLatestBlock(CHAIN);
+    const oldTarget = latest.timestamp - LOOKBACK_DAYS * SECONDS_PER_DAY;
+    old = await sdk.api.util.lookupBlock(oldTarget, { chain: CHAIN });
+  } catch (e) {
+    return 0;
```
0xkr3p: i think we should return null instead of 0 here so we don't ingest incorrect values because this impacts the overall APY 30d and can lead to skewed figures when it could just be rpc failure or other issue. Same applies elsewhere in the code

src/adaptors/btcd/index.js
```diff
@@ -0,0 +1,226 @@
...
+  return [
+    {
+      pool: `${SBTCD.toLowerCase()}-ethereum`,
+      chain: 'Ethereum',
+      project: 'btcd',
+      symbol: 'sBTCD',
+      tvlUsd: sBtcdTvlUsd,
+      apyBase: sBtcdApyBase,
+      underlyingTokens: [BTCD],
+      poolMeta,
```
0xkr3p, line 241: pls add `pricePerShare` field

0xkr3p (commented):
hi @z-j-lin, thanks for the PR, I left a couple of comments to address before we can merge. Pls apply the null change across the file where the apy value is set as 0 due to errors rather than the returns being 0%

src/adaptors/btcd/index.js
z-j-lin: if i add the pricepershare field there the check for allwoed field names test fails

0xkr3p (approved):
approved - thanks for the changes!

PR #2716, opened 2026-05-30
optimized O(n²) pool lookup to O(n) with Map in getPoolsBorrow

adityamane765:
getPoolsBorrow was calling pools.find() inside a .map() over lendBorrow -> resulting in O(n²) comparisons on every request. With thousands of pools on both sides this means millions of string comparisons per call. Replaced with a Map keyed by pool id built once before the loop, reducing the join to O(n).

## Summary by CodeRabbit

* **Refactor**
  * Improved performance of pool borrow data retrieval operations for faster response times.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/yield-server/pull/2716?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/api/controllers/enriched.js
+++ b/src/api/controllers/enriched.js
@@ -96,9 +96,10 @@ const getPoolsBorrow = async (req, res) => {
   const lendBorrow = data[1];
 
   // join supply side fields (all enriched fields) onto borrow object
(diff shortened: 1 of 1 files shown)


0xkr3p (approved):
hi @adityamane765, thanks for the PR - approved

PR #2717, opened 2026-05-30
parallelize block and token price fetches in utils

adityamane765:
Two sequential fetch patterns in utils.js replaced with Promise.all.
getBlocksByTime was awaiting each timestamp fetch one at a time inside a for loop, since each request is independent, all can be fired in parallel.

The token price chunking loop was also sequential -> each 50-token chunk waited for the previous one to complete before starting. Changed to build all chunks first then fetch in parallel with Promise.all, then merge results with Object.assign.

## Summary by CodeRabbit

* **Refactor**
  * Improved data-fetch performance: block height lookups and token price retrievals are now executed more concurrently and in batched requests, reducing latency and speeding up TVL and historical block queries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/yield-server/pull/2717?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/adaptors/utils.js
+++ b/src/adaptors/utils.js
@@ -119,7 +119,9 @@ const getBlocksByTime = async (timestamps, chainString) => {
   const chain = chainString === 'avalanche' ? 'avax' : chainString;
   const responses = await Promise.all(
     timestamps.map((timestamp) =>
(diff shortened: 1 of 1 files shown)


slasher125:
@adityamane765 hi, thanks for opening this PR, we pushed these changes via separate commits. 
closing this one here

PR #2719, opened 2026-06-02
Add Akka vault yield adapter

lkhoshnafs:
Akka AI Quant Vault on Hyperliquid L1. Fetches TVL and computes annualized APY from on-chain PnL history via the Hyperliquid API.

## Summary by CodeRabbit

* **New Features**
  * Added Hyperliquid "akka" vault integration exposing vault details in-app.
  * Shows total value locked (USD) for the vault and follower accounts.
  * Provides an estimated base APY with a fallback to reported APR when needed.
  * Returns a normalized single-pool view for this vault.

### DIFF
--- a/src/adaptors/akka/index.js
+++ b/src/adaptors/akka/index.js
@@ -1,0 +1,85 @@
+const axios = require('axios');
+const utils = require('../utils');
+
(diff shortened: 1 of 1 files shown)


0xkr3p:
Currently no TVL adapter, this is required before we can move forwards with a yield adapter

PR #2734, opened 2026-06-13
invalid uuid check is returning a plain object instead of sending HTTP response

adityamane765:
in getConfigPool the invalid uuid early return was doing return { status: 'invalid uuid parameter' } which is just a plain JS object from the async function so express alw ignores this and no HTTP response is ever sent, and the client connection hangs until timeout
fixed by calling res.status(400).json(...) instead

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid configuration identifiers now immediately return an HTTP 400 JSON response with a clear error message.
  * Configuration-related errors are now propagated to the centralized JSON error handler, ensuring consistent HTTP status codes and JSON {status, message} responses across failures.

### DIFF
--- a/src/api/app.js
+++ b/src/api/app.js
@@ -54,8 +54,8 @@ app.use('/', [yieldRoutes, config, median, perp, enriched, lsd, pools]);
 
 function errorHandler (err, req, res, next) {
   console.log(err)
-  res.status(500)
-  res.render('error', { error: err })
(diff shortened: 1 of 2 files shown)


0xkr3p:
thanks for your changes - i am closing this as we are migrating to new v2 api which has its own logic in place

PR #2745, opened 2026-06-18
Add OroSwap yield adapter

permapod-rsk:
## Summary by CodeRabbit

* **New Features**
  * Oroswap protocol integration: Added support for active mainnet pools with yield analytics, including TVL (USD), base APY, pool symbol and metadata, and underlying token denominations. The adapter gathers pool data via contract pagination, resolves pricing to compute TVL, and estimates APY from sampled recent swap fees, then formats results consistently and sorts by highest TVL.

### DIFF
--- a/src/adaptors/oroswap/index.js
+++ b/src/adaptors/oroswap/index.js
@@ -1,130 +1,510 @@
 const axios = require('axios');
+const BigNumber = require('bignumber.js');
 const utils = require('../utils');
 
-const API_URL = 'https://api-mainnet.oroswap.org/api/poolmanager/pools';
+const LCD_ENDPOINT = 'https://public-zigchain-lcd.numia.xyz';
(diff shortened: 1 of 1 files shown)


src/adaptors/oroswap/index.js
```diff
@@ -0,0 +1,511 @@
...
+      const current = index++;
+      results[current] = await mapper(items[current], current);
+    }
+  });
+
+  await Promise.all(workers);
+  return results;
+};
+
+const withRetry = async (fn, attempts = 3) => {
```
0xkr3p, line 109: use the withRetry from utils

0xkr3p (commented):
hi @permapod-rsk, thanks for the PR. nice PR - I think we should be doing more than a 100 sample size, particularly because stZIG has 10,000+ swaps each day. We should aim for at least 24 hours. We should try to be as accurate as possible within reason, otherwise we could be reporting a misleading figure where its better to not show the pool at all

PR #2753, opened 2026-06-22
Add XOXNO Lending yield adapter (Stellar + MultiversX)

mihaieremia:
XOXNO Lending is an overcollateralized lending protocol on MultiversX and Stellar.

This PR adds the yield pools for both chains, with supply and borrow APYs, LTV and available liquidity. On Stellar the rates live on shared hubs and the risk parameters on isolated spokes, so spokes are emitted as routing rows like aave-v4 does. MultiversX is a flat market model, so each market is a normal pool.

It also fixes two things on the Stellar side: available borrow was reporting the borrow cap instead of the liquidity actually in the pool, and the pool token is now set explicitly since positions are NFTs rather than receipt tokens.

The second commit fixes the pool table in adapter PR comments, which has been empty for every adapter since CI moved to Node 24. It is unrelated to XOXNO and easy to drop if you would rather take it separately, but it cannot be tested on its own: a PR touching only the workflow runs no adapter and posts no comment. The comment on this PR shows it working.


mihaieremia:
Update: the adapter now exports protocolId `8049`, matching the XOXNO Lending listing branch. The remaining protocol-slug CI check depends on merging the DefiLlama server listing for `xoxno-lending`; until `https://api.llama.fi/protocols` includes that slug, yield-server CI will keep failing at the protocol registry assertion.

0xkr3p:
converting this into a draft as no tvl adapter and appears to be on stellar testnet

mihaieremia:
> converting this into a draft as no tvl adapter and appears to be on stellar testnet

Thanks, will re convert once mainnet is deployed. This was a test commit to see if we pass the workflows 🫡

mihaieremia:
@0xkr3p the mainnet has been deployed, besides this PR I have few more https://github.com/DefiLlama/DefiLlama-Adapters/pull/20743 and https://github.com/DefiLlama/dimension-adapters/pull/9041 🙏

src/adaptors/xoxno-lending/index.js
```diff
@@ -0,0 +1,167 @@
...
+          symbol: market.symbol,
+          tvlUsd: cashUsd,
+          apyBase: decimalToPercent(market.supplyApy),
+          apyBaseBorrow: decimalToPercent(market.borrowApy),
+          totalSupplyUsd: Number(market.suppliedUsd ?? 0),
+          totalBorrowUsd: Number(market.borrowedUsd ?? 0),
+          availableBorrowUsd: cashUsd,
+          ltv: Number(market.ltv ?? 0),
+          borrowable: Boolean(market.borrowable),
+          underlyingTokens: [market.token],
```
0xkr3p: Add a normalizer beside decimalToPercent (after line 46), then use it at 72–73:

  // DefiLlama prices native EGLD under the null address; `elrond:EGLD` resolves
  // to nothing. Same fix as the TVL adapter.
  const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
  const priceToken = (chainSlug, token) =>
    chainSlug === 'elrond' && token === 'EGLD' ? NULL_ADDRESS : token;

src/adaptors/xoxno-lending/index.js
```diff
@@ -0,0 +1,167 @@
...
+
+    // Liquidity and APY live on the hub, so an asset on two hubs is two pools.
+    for (const market of hubMarkets) {
+      const tvlUsd = Number(market.tvlUsd ?? 0);
+      pools.push({
+        pool: `xoxno-lending-${chainSlug}-${market.poolId}`,
+        chain,
+        project: 'xoxno-lending',
+        symbol: market.symbol,
+        poolMeta: market.hubName || `Hub ${market.hubId}`,
```
0xkr3p: underlyingStateKey is absent everywhere. The hub poolId is already ${hubId}-${token}, so the key is free:

96 ```underlyingStateKey: `${market.hubId}-${market.token}`,```

Also on 131
 ```underlyingStateKey: `${spoke.hubId}-${spoke.token}`,```

  And line 136 — routeGroupKey is "1", "2", not namespaced by chain (aave-v4 uses the spoke address):

  136      ```routeGroupKey: `${chainSlug}-${spoke.spokeId}`,```

src/adaptors/xoxno-lending/index.js
```diff
@@ -0,0 +1,167 @@
...
+    const chainSlug = chain.toLowerCase();
+
+    if (config.model === 'flat') {
+      for (const market of Array.isArray(data.markets) ? data.markets : []) {
+        const cashUsd = Number(market.tvlCashUsd ?? 0);
+        pools.push({
+          pool: `xoxno-lending-${chainSlug}-${market.poolId}`,
+          chain,
+          project: 'xoxno-lending',
+          symbol: market.symbol,
```
0xkr3p: Symbols could be better formatted:

 Line 63          symbol: market.symbol,   // "StakedEGLD", "WrappedUSDC", "WrappedBTC", "OneDex", "Utrust"
 Line 95          symbol: market.symbol,   // Stellar: one is "CBSJZEIO5C7KC2SF" — a truncated address
 Line 129         symbol: spoke.symbol,

 On MultiversX the ticker is the ESDT prefix — market.token.split('-')[0] gives XEGLD / USDC / WBTC / WETH / ONE / UTK. Stellar needs a fallback when the API returns an address-shaped symbol.

src/adaptors/xoxno-lending/index.js
```diff
@@ -0,0 +1,167 @@
...
+}
+
+function decimalToPercent(value) {
+  return Number(value ?? 0) * 100;
+}
+
+async function apy() {
+  const pools = [];
+
+  for (const config of Object.values(CHAIN_CONFIGS)) {
```
0xkr3p: getExport throws on any non-2xx (lines 25–29), so a Stellar blip takes MultiversX — the entire $1.07M side — with it. Restructure to Promise.allSettled over CHAIN_CONFIGS

0xkr3p (commented):
hi @mihaieremia, thanks for the PR, I've left some comments, nice change on the commentResult.js but can we split out pls,as it needs to be separate for proper testing

mihaieremia:
> hi @mihaieremia, thanks for the PR, I've left some comments, nice change on the commentResult.js but can we split out pls,as it needs to be separate for proper testing

Should be good now.

I couldn't find anymore the repository where I could write the adaptor for coins pricing. I have it forked locally but meanwhile went private 😅 What is the current path to make a fix there for MvX and a LP pricing for Aquarius DEX which is already listed and has all the pools taken into account I just need to derive the share price as you already do for MultiversX.

0xkr3p (approved):
approved, thanks for the changes @mihaieremia 

our coins dev will look into pricing assets if they're used enough, typically we look at price assets that are over 100k mcap

PR #2764, opened 2026-06-29
Upgrade to loan router v2 and introduce escrow timelock

wegory:
## Summary by CodeRabbit

* **New Features**
  * Interest calculations now use live on-chain data for more accurate yield reporting.
  * Support expanded to include both PYUSD and USDAI as underlying tokens.

* **Bug Fixes**
  * Improved APY results by correctly combining fixed pool interest with loan accrual interest.
  * Added a safe early return when redemption value is zero to avoid empty or misleading outputs.


0xkr3p (approved):
approved thanks

PR #2769, opened 2026-07-01
canto-lending: report available liquidity as TVL

permapod-rsk:
## Summary

  This updates the `canto-lending` adapter so that the displayed `tvlUsd` represents the liquidity currently available in the market, while keeping the broader lending accounting fields (`totalSupplyUsd`, `totalBorrowUsd`,
  `availableBorrowUsd`) explicit.

  ## Motivation

  For lending markets, the value shown as TVL should reflect the liquidity that is actually available in the pool/reserve, consistent with the way Aave v3 exposes reserve liquidity and with the intended lending-market display
  semantics.

  Previously, `canto-lending` computed:

  `tvlUsd = totalSupplyUsd - totalBorrowUsd`

  Since `totalSupplyUsd` was derived from `cash + borrows - reserves`, this effectively reduced to available cash minus reserves. That made the displayed TVL less directly tied to the onchain available liquidity value.

  ## What changed

  - `tvlUsd` is now calculated directly from the market's onchain `getCash()` value:
    - `tvlUsd = getCash * underlyingPrice`
  - `availableBorrowUsd` is set to the same available liquidity value.
  - `totalSupplyUsd` and `totalBorrowUsd` are still exported separately for lending analytics.
  - `apyBase` and `apyBaseBorrow` continue to come from onchain `supplyRatePerBlock` and `borrowRatePerBlock`, now using a Compound-style compounded APY calculation.
  - Reward APY calculation was split into a dedicated helper so base APY and reward APY are not mixed in the same function.

  ## Expected result

  For Canto Lending markets, DefiLlama will now display TVL as the amount of liquidity currently available to withdraw/borrow from the market, while preserving the full lending context through:

  - `totalSupplyUsd`
  - `totalBorrowUsd`
  - `availableBorrowUsd`
  - `apyBase`
  - `apyBaseBorrow`

## Summary by CodeRabbit

* **New Features**
  * Added support for Valdora Finance staking APY and TVL data on ZIGChain.
  * Improved Canto lending APY and reward yield calculations for more accurate returns.
* **Bug Fixes**
  * Updated available liquidity and reward APY calculations to better handle invalid or non-positive values.

### DIFF
--- a/src/adaptors/canto-lending/index.js
+++ b/src/adaptors/canto-lending/index.js
@@ -215,13 +215,20 @@ const getPrices = async (chain, addresses) => {
   }, {});
 };
 
-function calculateApy(rate, price = 1, tvl = 1) {
-  // supply rate per block * number of blocks per year
+function calculateApy(rate) {
   const BLOCK_TIME = 6;
   const YEARLY_BLOCKS = (365 * 24 * 60 * 60) / BLOCK_TIME;
-  const safeTvl = tvl === 0 ? 1 : tvl;
-  const apy = (((rate / 1e18) * YEARLY_BLOCKS * price) / safeTvl) * 100;
-  return apy;
+
+  return (Math.pow(rate / 1e18 + 1, YEARLY_BLOCKS) - 1) * 100;
+}
+
+function calculateRewardApy(speed, rewardPrice, totalUsd) {
+  const BLOCK_TIME = 6;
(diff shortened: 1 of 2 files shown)


permapod-rsk:
FYI the front end is dead but the protocol is still alive.

0xkr3p:
> FYI the front end is dead but the protocol is still alive.

How does the user deposit? We should probably exclude this adapter, if a user is unable to deposit

PR #2773, opened 2026-07-02
Merkl balancerv3

franzns:
Redo of https://github.com/DefiLlama/yield-server/pull/2630

Using the internal `merkl-additional-reward.js` to fetch direct campaigns. Additionally handle forwarded campaigns for boosted pools. 

They now show proper reward tokens for each merkl campaign.


0xkr3p (approved):
thanks for the pr @franzns - approved

PR #2778, opened 2026-07-05
Create index.js

yurac777:
## Summary by CodeRabbit

* **New Features**
  * Added support for a new Alm Vault listing with live on-chain TVL calculation.
  * Displayed fixed base and reward APY values plus reward token information.
  * Included the vault’s public website link in the listing.


0xkr3p (commented):
hi @yurac777, thanks for the pr! A couple of things required before we can proceed:
- A TVL adapter is required
- apy base/reward should not be hardcoded and should be calculated from on chain data
- tvl should not fallback to hardcoded 50000

PR #2780, opened 2026-07-07
Midas meta tags 

waynebruce0x:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token symbol display for additional pools (including assets previously missing denomination info), so assets now show the correct denomination more consistently.
  * Updated pool labeling to derive symbols from denomination when available, with a clearer fallback when it isn’t, reducing ambiguous or incorrect labels.


0xkr3p (approved):
LGTM

PR #2784, opened 2026-07-07
fix(marginfi-lending): read banks on-chain, snapshot S3 bucket is gone

0xshubhs:
## Problem

The adapter fetched a pre-generated snapshot from `https://marignfi-pools-snapshot.s3.amazonaws.com/snapshot.json`. That bucket no longer exists, the request returns `404 NoSuchBucket` (XML), so the adapter throws `Unexpected token '<' ... is not valid JSON` on every run. marginfi (~$34M TVL, protocolId 2570) has been missing from yields since.

There is no replacement REST endpoint; the marginfi app computes rates client-side from RPC. So this rewrites the adapter to read everything on-chain:

## Approach

- Bank list + symbols from marginfi's public bank metadata cache (`storage.googleapis.com/mrgn-public/mrgn-bank-metadata-cache.json`, same file their frontend uses)
- Bank accounts via a single `getMultipleAccounts` call on mainnet RPC, decoded with offsets derived from the marginfi-v2 IDL 0.1.7 (bytemuck/repr(C) with explicit padding, I80F48 fixed-point fields)
- Lending/borrowing rates read from the program's own `BankCache` (spot APRs the program computes on every interaction, u32-scaled where u32::MAX = 1000%), with the legacy interest-rate-curve as fallback for banks whose cache is untouched. APR -> APY with hourly compounding, matching the marginfi UI
- Prices from coins.llama.fi (keys built case-sensitively, Solana mints don't survive lowercasing)
- Only `Operational` banks are included; paused/reduce-only are skipped

Pool ids are the bank addresses. The old S3 snapshot format isn't archived anywhere I could find, so I couldn't confirm the previous pool id convention.

## Output

42 pools, ~$32M combined tvlUsd (protocol TVL on DefiLlama is $34.5M, remainder is staked-collateral/arena banks not in the main cache):

| symbol | tvlUsd | totalSupplyUsd | apyBase | apyBaseBorrow | ltv |
|---|---|---|---|---|---|
| JitoSOL | $6.25M | $6.99M | 0.14% | 2.53% | 0.75 |
| mSOL | $4.67M | $5.24M | 0.15% | 2.60% | 0.65 |
| jupSOL | $2.70M | $2.93M | 0.08% | 2.15% | 0.65 |
| Bonk | $2.62M | $2.63M | 0.00% | 1.10% | 0.50 |
| WBTC | $1.20M | $1.36M | 0.17% | 2.57% | 0.50 |

Decoded rates cross-checked against the on-chain cache at three utilization levels (10.6%, 91.8%, 99.1%), cache and decode agree.

`npm run test --adapter=marginfi-lending`: **392 passed, 392 total**

## Summary by CodeRabbit

* **New Features**
  * Marginfi lending pool metrics now derive from live on-chain bank state rather than a static snapshot.
  * Pool totals in USD (including TVL, supply/borrow values, LTV, and APY) are computed using current token pricing and on-chain interest/curve inputs.
* **Bug Fixes**
  * Added fallback APR/APY computation when cached rate data isn’t available.
  * Improved handling of invalid results by filtering out non-finite pool values.

### DIFF
--- a/src/adaptors/marginfi-lending/index.js
+++ b/src/adaptors/marginfi-lending/index.js
@@ -34,6 +34,10 @@ const U32_MAX = 4294967295;
 // BankCache rates are u32-scaled APRs where u32::MAX represents 1000%
 const CACHE_RATE_SCALE = 10;
 const HOURS_PER_YEAR = 365.25 * 24;
+// Highest offset read is cacheBorrowingRate (u32)
+const MIN_BANK_DATA_SIZE = OFFSET.cacheBorrowingRate + 4;
+const RPC_TIMEOUT_MS = 30000;
+const RPC_RETRIES = 3;
 
 // I80F48 fixed point: 16 bytes little-endian, two's complement, 48 fraction bits
 const readI80F48 = (buf, offset) => {
@@ -82,8 +86,13 @@ const computeRates = (bank, utilization) => {
   return { lendingApr, borrowingApr };
 };
 
-const decodeBank = (data) => {
+const decodeBank = (data, bankAddress) => {
   const buf = Buffer.from(data, 'base64');
+  if (buf.length < MIN_BANK_DATA_SIZE) {
+    throw new Error(
+      `Bank ${bankAddress}: account data too short (${buf.length} < ${MIN_BANK_DATA_SIZE} bytes)`
+    );
+  }
   return {
     mint: buf.subarray(OFFSET.mint, OFFSET.mint + 32),
     mintDecimals: buf.readUInt8(OFFSET.mintDecimals),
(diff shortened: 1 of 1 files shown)


0xshubhs:
Addressed the review comments in d821e29:

- **RPC timeout + retry (CodeRabbit):** RPC calls now go through a small wrapper with a 30s timeout and up to 3 retries with exponential backoff on 429 / 5xx / timeouts.
- **decodeBank guards (Copilot):** added a minimum-length check before decoding, errors now carry the bank address, and undecodable accounts (e.g. stale metadata entries) are skipped with a log instead of failing the whole run.
- **$0-TVL pools with non-zero APY (CodeRabbit nitpick):** banks with no coins.llama.fi price are now skipped entirely.

`npm run test --adapter=marginfi-lending`: 392 passed, 43 pools.

0xkr3p:
hi @0xshubhs, this is awesome work - thank you for your effort. Unfortunately the marginfi-lending adapter is not in use and instead project 0 is being used, feel free to double check we are capturing everything

0xshubhs:
Hey @0xkr3p , thanks for letting me know . Got it

PR #2794, opened 2026-07-08
fix: indigo pools all had tvlUsd 0, ada price endpoint is broken

0xshubhs:
indigo's stability pool APRs were still coming through fine, but every pool had tvlUsd 0, which got it swept into the exclusion list at the end of May.

Root cause: the adapter converts the pools' ADA-denominated TVL to USD via `analytics.indigoprotocol.io/api/price?from=ADA&to=USD`, and that endpoint now returns HTTP 500. The catch swallows it and returns 0, so every pool's TVL multiplies out to 0. The rest of the analytics API (stability-pools, apr, asset analytics) works fine.

Fix: take the ADA price from coins.llama.fi (`coingecko:cardano`) through the existing `getPriceApiData` helper instead, and remove the exclusion entry.

Local run: 45/45 tests, 6 pools, ~$4.1M total (iUSD $2.82M, iBTC $1.10M, iETH $194k + iSOL/iJPY/iEUR dust), matches the ADA totals on their analytics API times the current ADA price, and the protocol page. APRs are passed through from indigo's own apr endpoint, unchanged by this PR.

### DIFF
--- a/src/adaptors/indigo/index.js
+++ b/src/adaptors/indigo/index.js
@@ -1,4 +1,5 @@
 const { get, post } = require('./http-helper')
+const utils = require('../utils')
 
 const CARDANO_COINGECKO = {
   iUSD: 'coingecko:iusd',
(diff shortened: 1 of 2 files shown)


0xkr3p (approved):
approved thanks for the PR @0xshubhs

PR #2795, opened 2026-07-08
fix: saucerswap-v1 broken since api.saucerswap.finance went key-only

0xshubhs:
saucerswap-v1 has been dead and excluded because api.saucerswap.finance now returns 401 Unauthorized on every endpoint the adapter uses (/farms, /pools, /pools/5day-avg); they made the API key-only. The protocol itself is fine (~$8.7M TVL on Hedera, 22 V1 farms still have live SAUCE emissions).

The endpoints work unchanged with an `x-api-key` header and the response shapes are identical, so this just threads `SAUCERSWAP_API_KEY` through the three getData calls, wires it into env.js/serverless.yml like the other keys, and removes the exclusion entry.

You already have a working key for this, dimension-adapters ships `SAUCERSWAP_API_KEY` in helpers/env.ts, and I ran the harness with that same key: 516/516 tests, 73 pools with sane values (SAUCE-HBAR $1.14M @ 0.47% base + 5.81% reward, USDC-HBAR $552k @ 4.60% + 3.90%, etc). It just needs to be present in this repo's deployment env too.

CI bot run will fail with 401 since the fork doesn't have the key, same situation as the Allium-based adapters in dimension-adapters.

### DIFF
--- a/env.js
+++ b/env.js
@@ -25,6 +25,7 @@ module.exports = {
   // DB
   DATABASE_URL: process.env.DATABASE_URL,
   OSMOSIS_API_KEY: process.env.OSMOSIS_API_KEY,
+  SAUCERSWAP_API_KEY: process.env.SAUCERSWAP_API_KEY,
   DUNE_API_KEY: process.env.DUNE_API_KEY,
   HYPERLIQUID_RPC: process.env.HYPERLIQUID_RPC,
   PLASMA_RPC: process.env.PLASMA_RPC,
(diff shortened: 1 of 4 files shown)


0xshubhs:
The failing test is the expected 401 — the CI run doesn't have SAUCERSWAP_API_KEY (fork PRs don't get secrets). With the key set it's 516/516 tests and 73 pools; the key from dimension-adapters helpers/env.ts works.

PR #2735, opened 2026-06-13
feat(t3tris-finance): add T3tris Finance APY adapter

t3trominos:
## Protocol info

- **Description**: T3tris is a zero fee, permissionless vault infrastructure.
- **Website**: https://t3tris.finance
- **Category**: Onchain Capital Allocator

## Methodology descriptions

- **TVL**: Total value of assets held across all verified T3tris vaults, including assets deployed in strategies, pending deposits, and claimable redemptions.
- **Fees**: Total yield generated by vault strategies, including depositor yield, curator fees and T3tris treasury profit.
- **Revenue**: Earnings sent to the T3tris treasury. Third-party services are also T3tris revenue but are not observable on-chain.

---

## What this PR adds

A new APY adapter for T3tris Finance vaults on Arbitrum.

### Data sources
- **Vault discovery**: ecosystem API (`ecosystem.t3tris.finance/vaults`) — only verified, non-blacklisted, public vaults
- **APY**: annualized from on-chain oracle PPS over a single fixed 7-day window, anchored at the vault's last oracle valuation timestamp
- **TVL**: `vault.totalAssets()` × asset price (USD)
- **pricePerShare**: oracle `getLastSavedPricePerShare` (immune to settlement-cycle artifacts)

### APY formula
```
APY = (PPS_current / PPS_(LVT−7d) − 1) / 7 × 365 × 100
```
Single fixed 7-day window, consistent across all vaults. The historical price-per-share is read 7 days before the vault's last oracle valuation timestamp (last NAV) and the current PPS is the value saved at that NAV, so the realized rate is held rather than decaying as flat days accumulate.

PR #2806, opened 2026-07-11
feat(t3tris-finance): surface entry & exit fees in poolMeta

t3trominos:
## What

The T3tris Finance adaptor already reads all four vault fees on-chain (performance, management, entry, exit) but only surfaced perf & mgmt in `poolMeta`. This PR adds the entry and exit fees so the full fee schedule is visible in the pool metadata.

## Details

- **No new on-chain calls** — `getEntryFee`/`getExitFee` were already fetched and stored per vault; only the display string changed.
- `poolMeta` is now assembled from a parts list: the performance fee is always shown, and management, entry, and exit are appended only when non-zero (so vaults that don't charge a given fee stay uncluttered).
- Example output: `perf 10.0%, mgmt 2.0%, entry 0.5%, exit 0.5%`

Follow-up to #2735.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated displayed pool fee information to consistently include vault performance fees.
  * Improved the fee summary formatting to show management, entry, and exit fee components only when their rates are non-zero.
  * Refined percentage display to use rounded values without trailing zeros.

### DIFF
--- a/src/adaptors/t3tris-finance/index.js
+++ b/src/adaptors/t3tris-finance/index.js
@@ -404,14 +404,15 @@ const main = async () => {
         // Fee metadata: always show the performance fee, then append the
         // management, entry, and exit fees whenever they are non-zero (kept
         // conditional so vaults that don't charge them aren't cluttered with 0%).
-        const feeParts = [`perf ${vault.perfFeePct.toFixed(1)}%`];
-        if (vault.mgmtFeePct > 0)
-          feeParts.push(`mgmt ${vault.mgmtFeePct.toFixed(1)}%`);
-        if (vault.entryFeePct > 0)
-          feeParts.push(`entry ${vault.entryFeePct.toFixed(1)}%`);
-        if (vault.exitFeePct > 0)
-          feeParts.push(`exit ${vault.exitFeePct.toFixed(1)}%`);
-        const poolMeta = feeParts.join(', ');
+        // Compact labels (in/out) and a '/' separator keep the string short;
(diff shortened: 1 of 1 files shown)


0xkr3p (commented):
hi @t3trominos, this pool meta is too long to be output properly... `perf 10.0%, mgmt 2.0%, entry 0.5%, exit 0.5%`. We discussed this on the previous PR

t3trominos:
> hi @t3trominos, this pool meta is too long to be output properly... `perf 10.0%, mgmt 2.0%, entry 0.5%, exit 0.5%`. We discussed this on the previous PR

You're right the four-fee string was too long. Rather than drop the fees, I reworked the format: 
- compact labels + / separator
- rates with trailing zeros stripped (2-decimal precision kept, so fine rates like 0.05% aren't rounded).
- Real vaults now render like perf 20% and perf 20%/mgmt 1%/out 0.4% (~25 chars).
- The `perf 10.0%, mgmt 2.0%, entry 0.5%, exit 0.5%` case you flagged becomes `perf 10%/mgmt 2%/in 0.5%/out 0.5%`.

Let me know if this works for you, happy to adjust.

0xkr3p:
closing this, previously discussed on #2735. Keep as perf mgmt. The user can see entry / exit fees when viewing t3tris site pool page

t3trominos:
> closing this, previously discussed on #2735. Keep as perf mgmt. The user can see entry / exit fees when viewing t3tris site pool page

If ever you want to give it a try, here is how it would render:
image
I think an entry/exit fee is critical and should be rendered if possible.

PR #2814, opened 2026-07-13
Add Robinhood Chain to T3tris Finance APY adapter

t3trominos:
T3tris Finance is now deployed on Robinhood Chain (4663). Adds `robinhood` to the APY adapter's chain list — vaults are auto-discovered from the T3tris ecosystem API per chainId (same curation gate: verified & non-blacklisted). The chain's @defillama/sdk provider is registered as `robinhoodchain`, so on-chain calls map the slug accordingly while the pool chain/URL keep the canonical `robinhood` slug (formatChain -> 'Robinhood Chain'). Kept separate from #2806 (poolMeta) to isolate concerns.

## Summary by CodeRabbit

* **New Features**
  * Added support for Robinhood chain data in the T3tris Finance integration.
  * Enabled historical and current oracle price retrieval, block resolution, and asset pricing for the new chain.

### DIFF
--- a/src/adaptors/utils.js
+++ b/src/adaptors/utils.js
@@ -197,6 +197,7 @@ exports.formatChain = (chain) => {
   if (chain && chain.toLowerCase() === 'optimism') return 'OP Mainnet';
   if (chain && chain.toLowerCase() === 'ton') return 'TON';
   if (chain && chain.toLowerCase() === 'zklink') return 'zkLink Nova';
+  if (chain && chain.toLowerCase() === 'robinhood') return 'Robinhood Chain';
 
   return chain.charAt(0).toUpperCase() + chain.slice(1);
 };
--- a/src/adaptors/utils.js
+++ b/src/adaptors/utils.js
@@ -47,13 +47,11 @@ const getStarknetRpc = () => {
   return process.env.STARKNET_RPC;
 };
 
-const rateLimitedStarknetCall =
-  (fn) =>
-  (...args) => {
-    const promise = pendingStarknetCall.then(() => fn(...args));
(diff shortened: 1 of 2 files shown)


src/adaptors/utils.js
```diff
@@ -199,6 +197,7 @@ exports.formatChain = (chain) => {
   if (chain && chain.toLowerCase() === 'optimism') return 'OP Mainnet';
   if (chain && chain.toLowerCase() === 'ton') return 'TON';
   if (chain && chain.toLowerCase() === 'zklink') return 'zkLink Nova';
+  if (chain && chain.toLowerCase() === 'robinhood') return 'Robinhood Chain';
```
0xkr3p: this has been implemented on line 182, pls remove duplicate addition

t3trominos: Sorry! It's done.

0xkr3p:
hi @t3trominos - `getPrices` and `getBlockNumber` should also be using the `robinhoodchain` mapping I believe as currently that uses robinhood

t3trominos:
> hi @t3trominos - `getPrices` and `getBlockNumber` should also be using the `robinhoodchain` mapping I believe as currently that uses robinhood

Good catch, thanks @0xkr3p done in the latest commit.
- Wrapped both `getBlockNumber` (the `coins.llama.fi/block/{chain}` lookup) and `utils.getPrices` with the same `sdkChain` mapping, so all five backend lookups now use `robinhoodchain`. Only the pool `chain`/URL keep the canonical `robinhood` slug via `formatChain`.

0xkr3p (approved):

PR #2818, opened 2026-07-15
Add Tristero AUSD yield adapter

ggarza5:
## Summary

- Add the Ethereum AUSD Tristero margin-lending vault to the yields adapter set.
- Derive APY from the vault’s on-chain `ratePerSecond` asset parameter using standard daily compounding.
- Derive TVL from the vault’s `getTVOL(AUSD)` balance and AUSD price.

## Validation

- `git diff --check` passes.
- The adapter was manually exercised against live Ethereum contract data and returned TVL and APY values.

Arbitrum and Base deployments are intentionally excluded because they use USDC and need separate confirmation of their lender-yield treatment.

## Summary by CodeRabbit

* **New Features**
  * Added Tristero’s AUSD margin-lending vault to yield tracking.
  * Added reporting for vault total value and estimated APY, including a share price metric.
  * Added a direct link to the Tristero app for more information.

### DIFF
--- a/src/adaptors/tristero-margin/index.js
+++ b/src/adaptors/tristero-margin/index.js
@@ -1,0 +1,89 @@
+// DeFiLlama yield adapter for the Tristero AUSD margin-lending vault.
+//
+// The vault is not an ERC-4626 share vault. Its on-chain accounting exposes
+// getTVOL(asset) for TVL and assets(tokenId).ratePerSecond for the configured
+// lender rate. APY is derived from that explicit contract rate; TVOL is never
+// used as a yield proxy.
+//
+// This deliberately lists only the Ethereum AUSD deployment. The Arbitrum and
+// Base deployments use USDC and should be added separately once their lender
+// yield treatment is confirmed.
+
+const sdk = require('@defillama/sdk');
+const utils = require('../utils');
+
+const CHAIN = 'ethereum';
+const VAULT = '0xB49781E8c39c75f413C1178f395bF68b0BEE8d00';
+const AUSD = '0x00000000eFE302BEAA2b3e6e1b18d08D69a9012a';
+const SECONDS_PER_YEAR = 365 * 24 * 60 * 60;
+const AUSD_ID = BigInt(AUSD).toString();
+
+const getTVOL = async (block) => {
+  const result = await sdk.api.abi.call({
+    target: VAULT,
+    abi: 'function getTVOL(address _token) view returns (uint256)',
+    params: [AUSD],
+    chain: CHAIN,
+    ...(block ? { block } : {}),
+  });
+
+  return BigInt(result.output);
+};
+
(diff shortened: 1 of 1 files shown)


src/adaptors/tristero-margin/index.js
```diff
@@ -0,0 +1,89 @@
...
+    pool: `${VAULT}-${CHAIN}`.toLowerCase(),
+    chain: utils.formatChain(CHAIN),
+    project: 'tristero-margin',
+    symbol: 'AUSD',
+    tvlUsd,
+    apyBase,
+    underlyingTokens: [AUSD],
+    poolMeta: 'AUSD margin lending vault',
+    url: 'https://app.tristero.com',
+    token: null,
```
0xkr3p, line 89: does the user receive a receipt token for depositing? if so this should be added here rather than null

src/adaptors/tristero-margin/index.js
```diff
@@ -0,0 +1,89 @@
...
+  return [{
+    pool: `${VAULT}-${CHAIN}`.toLowerCase(),
+    chain: utils.formatChain(CHAIN),
+    project: 'tristero-margin',
+    symbol: 'AUSD',
+    tvlUsd,
+    apyBase,
+    underlyingTokens: [AUSD],
+    poolMeta: 'AUSD margin lending vault',
+    url: 'https://app.tristero.com',
```
0xkr3p, line 86: add pricePerShare field

src/adaptors/tristero-margin/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+  return [{
+    pool: `${VAULT}-${CHAIN}`.toLowerCase(),
+    chain: utils.formatChain(CHAIN),
+    project: 'tristero-margin',
+    symbol: 'AUSD',
+    tvlUsd,
+    apyBase,
+    underlyingTokens: [AUSD],
+    poolMeta: 'AUSD margin lending vault',
```
0xkr3p: I don't think this pool meta adds anything valuable, can remove

0xkr3p (commented):
thanks for the PR @ggarza5, a few minor tweaks before we can move forwards

src/adaptors/tristero-margin/index.js
ggarza5: Thanks — I checked the dapp implementation and the verified MarginVault source. Deposits mint ERC6909 shares on the MarginVault using the asset ID uint256(uint160(AUSD)); there is no standalone ERC20 receipt-token address. I therefore kept token: null to avoid misrepresenting holder analytics. I added pricePerShare using readValue(AUSD, 1e18), which returns the current time/rate-accrued index, and removed poolMeta. Commit: f3bf91f.

0xkr3p (approved):

PR #2829, opened 2026-07-17
remove isprivate condition from yield server

burhanuddin03:
Removed isPrivate condition because it is used by our FE to enable/disable deposits on our UI based on a whitelist we set onchain for certain vaults, this doesn't constitute an internal vault but rather it is a mechanism that some managers use to have a better control of inflows from LPs that want to deposit. It also doesn't hinder the operations of the vault, the strategies are still deployed across multiple chains/protocols and accrue yield in the same way as other vaults.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated vault metadata retrieval to use the current API endpoint.
  * Private vaults can now be included in supported pool listings and related APY and TVL calculations when they meet eligibility requirements.

### DIFF
--- a/src/adaptors/ember-protocol/index.js
+++ b/src/adaptors/ember-protocol/index.js
@@ -5,7 +5,7 @@ const utils = require('../utils');
 const PROJECT = 'ember-protocol';
 const BASE_URL = 'https://ember.so/';
 const EARN_URL = 'https://ember.so/earn';
-const VAULTS_URL = 'https://vaults.api.sui-prod.bluefin.io/api/v2/vaults';
+const VAULTS_URL = 'https://vaults.api.prod.ember.so/api/v2/vaults';
 
 const poolUrl = (receiptSymbol) =>
   receiptSymbol ? `${EARN_URL}/${receiptSymbol}` : BASE_URL;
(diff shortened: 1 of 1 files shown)


0xkr3p:
> isPrivate condition because it is used by our FE to enable/disable deposits 

hi @burhanuddin03, thanks for the PR we only want to show yields where deposits are enabled.. if a user can't deposit into the vault us displaying to them is not good UX. Happy to move ahead with the api change, revert the isPrivate change.

burhanuddin03:
> > isPrivate condition because it is used by our FE to enable/disable deposits 
> 
> 
> 
> hi @burhanuddin03, thanks for the PR we only want to show yields where deposits are enabled.. if a user can't deposit into the vault us displaying to them is not good UX. Happy to move ahead with the api change, revert the isPrivate change.


Hey @0xkr3p  understand your point but users are indeed able to deposit into these vaults, they just need to request an allocation for the whitelist but the process is smooth and we've been able to onboard multiple users with it.

We strongly believe it is more valuable showing these vaults than hiding them since they are good opportunities for users to invest in some of the highest yielding vaults in the platform, while also getting familiar with the vault manager during the onboarding process for added trust.

0xkr3p (approved):

PR #2843, opened 2026-07-23
unexclude yieldfi, its apy feed is live again

0xshubhs:
yieldfi got excluded on jul 12 - fair call at the time, their ctrl.yield.fi apy feed had been serving 0 since jul 9 so the adapter (fixed in #2792 the same day) had nothing to report. the feed came back today with real numbers (yusd 9.03%, vyusd 9.13%, 383 days of history on the endpoint), so this removes the exclude entry, same as the amnis one in #2793.

also made the adapter fetch each symbol's apy once up front instead of once per (symbol, chain) - it was hitting the same endpoint ~30 times in parallel and random pools dropped whenever a duplicate call flaked (i was getting 18 pools on one run and 30 on the next). with the change the count is stable.

test: 215 passed, 35 pools, 9.9M total across ethereum/saga/base/arbitrum + smaller chains, ran it twice with matching results

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YieldFi pool data processing for more consistent APY and TVL results.
  * YieldFi pool APY handling was updated to avoid missing or mismatched APY values in listings.
* **Behavior Changes**
  * YieldFi is no longer excluded by a static safety list and will only be filtered out when flagged by active safety checks.

### DIFF
--- a/src/adaptors/yieldfi/index.js
+++ b/src/adaptors/yieldfi/index.js
@@ -282,9 +282,10 @@ const poolsFunction = async () => {
   // duplicate call flaked
   const symbols = ['yUSD', 'vyUSD', 'yETH', 'vyETH', 'yBTC', 'vyBTC', 'yPrism', 'yHLP', 'yValos', 'yPYMN'];
   const apys = {};
-  for (const symbol of symbols) {
-    apys[symbol] = await fetchLatestAPY(symbol);
-  }
+  const apyValues = await Promise.all(symbols.map((symbol) => fetchLatestAPY(symbol)));
+  symbols.forEach((symbol, i) => {
+    apys[symbol] = apyValues[i];
+  });
 
   // Process all chains in parallel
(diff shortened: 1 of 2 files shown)


0xshubhs:
moved the apy fetches to Promise.all, each symbol still requested once. same result on the tests, 35 pools twice

0xshubhs:
@coderabbitai review

0xkr3p:
hi @0xshubhs, thanks for looking into this. unfortunately gonna have to close this PR because their frontend doesn't work, they've deleted most of their tweets and are currently in a remediation with SAGA: `"Mint and Redeem operations for yUSD and vyUSD are paused until we seek remediation from SAGA as per the contractual agreement with them."` I think its best we leave this as excluded for now

check their frontend: https://v2.yield.fi/vyusd

0xshubhs:
NTW got it 🫡

PR #2857, opened 2026-07-30
add zensats adapter

NateBrune:
Adds a yields adapter for ZenSats — ERC-4626 "Zenji" leverage-loop vaults that deposit a collateral asset, borrow USDT against it on LlamaLend, and loop the proceeds back in.

Protocol slug `zensats` (id 7621) confirmed via https://api.llama.fi/protocols. Already listed on the TVL page.

### Pools (2)

| Chain | Pools | TVL |
|---|---|---|
| Ethereum | 2 | $11,114 |

$11,114 total; 1 pool clears the $10k display floor. Both vaults launched 2026-06-13, so balances are still small.

| Chain | Symbol | Market | TVL | apyBase | pricePerShare |
|---|---|---|---|---|---|
| Ethereum | WBTC | USDT/crvUSD LlamaLend StakeDAO | $10,074 | 3.95% | 1.001482 |
| Ethereum | wstETH | USDT/crvUSD LlamaLend StakeDAO | $1,040 | 3.69% | 1.001522 |

### Data source

Entirely on-chain, no API. `totalAssets()` for TVL and `convertToAssets(1 share)` for share price, batched via `sdk.api.abi.multiCall` with `permitFailure: true`. Historical share prices use blocks resolved through `utils.getPriceApiData('/block/ethereum/{ts}')`, underlying prices via `utils.getPrices`.

Both vaults are standard ERC-4626. Verified on-chain that `asset()` matches the configured underlying, that the asset symbol matches the configured `symbol`, and that share decimals equal asset decimals (8 for WBTC, 18 for wstETH) — the `convertToAssets` scaling depends on the last one.

### APY methodology

`apyBase` is the annualized growth of `convertToAssets(1 share)` over a **7 day** window, and the same figure is emitted as `apyBase7d`. The 24h figure is computed as well, but used only as a fallback for when the 7d reading is unusable (vault younger than the window, or a failed archive call). This follows the `geth` adapter.

Not the 24h window, because these are leverage loops: the share price is collateral-denominated while the debt is stablecoin-denominated, so it marks to market with the collateral/debt ratio and is not monotonic. The 24h readings measured +2.32% and −2.38% while 7d/14d/30d growth was positive throughout, and a negative `apyBase` is clamped to 0 downstream — so the wstETH pool would publish 0% instead of 3.69%.

Harvests are discrete, so a single harvest inside the window visibly moves the number at this size. Both vaults were harvested during testing: WBTC went 3.26% → 3.95% and wstETH 2.94% → 3.69%, each off one ~0.014% step in share price. The 7d window damps this considerably relative to 24h but doesn't eliminate it.

No `apyReward` or `rewardTokens` — these vaults pay no incentive tokens, and no points or pre-TGE incentives are included.

### Three pools deliberately excluded

The TVL registry entry also lists three pmUSD/crvUSD StakeDAO vaults, omitted here for now:

| Vault | Asset | totalAssets |
|---|---|---|
| `0x617A6877…` | WBTC | 0 |
| `0xbaEc8343…` | wstETH | 0 |
| `0x7d5281D5…` | XAUt | 0.0359 (~$145) |

Two are empty. They are older contracts and they are deprecated.

Neither included pool id is claimed by another projellama.fi/pools`, both return no existing match.

### Test

```
npm run test --adapter=zensats
PASS ./test.js
Tests:       23 passed, 23 total
```

## Summary by CodeRabbit

* **New Features**
  * Added ZenSats ERC-4626 “Zenji” vault support for WBTC and wstETH.
  * Calculates TVL (USD), price-per-share, and an annualized APY using a 7-day on-chain share-to-asset comparison.
  * Includes ZenSats adapter metadata and links for discoverability.
* **Bug Fixes**
  * Improved result robustness by discarding vaults when required inputs are missing, conversions are invalid, or `totalAssets` is non-positive or failed.

### DIFF
--- a/src/adaptors/zensats/index.js
+++ b/src/adaptors/zensats/index.js
@@ -77,8 +77,15 @@ const apy = async (timestamp) => {
 
   return VAULTS.map((v, i) => {
     const price = pricesByAddress[v.underlyingToken.toLowerCase()];
-    const total = Number(totalAssets.output[i]?.output);
-    if (!price || !Number.isFinite(total)) return null;
+    const totalRes = totalAssets.output[i];
+    const total = Number(totalRes?.output);
+    if (
+      !price ||
+      totalRes?.success === false ||
+      !Number.isFinite(total) ||
+      total <= 0
+    )
+      return null;
 
     const unit = 10 ** v.decimals;
     const current = Number(ppsNow.output[i]?.output);
--- a/src/adaptors/zensats/index.js
+++ b/src/adaptors/zensats/index.js
@@ -93,12 +93,17 @@ const apy = async (timestamp) => {
 
     const unit = 10 ** v.decimals;
     const current = Number(ppsNow.output[i]?.output);
+    // Without a current share price there is no APY to report. Drop the pool
+    // rather than publishing apyBase: 0, which would overwrite the last good
+    // value with a rate the vault never paid.
+    if (!(current > 0)) return null;
 
-    // Annualized share-price growth over `days`, or null when either endpoint
-    // is missing — a vault deployed inside the window, or a failed archive call.
+    // Annualized share-price growth over `days`, or null when the historical
+    // endpoint is missing — a vault deployed inside the window, or a failed
+    // archive call. Then apyBase: 0 is correct: no growth has been observed.
     const growth = (pastRes, days) => {
       const past = Number(pastRes.output[i]?.output);
-      if (!(current > 0) || !(past > 0)) return null;
+      if (!(past > 0)) return null;
       const value = ((current / past) ** (365 / days) - 1) * 100;
       return Number.isFinite(value) ? value : null;
     };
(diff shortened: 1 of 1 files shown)


src/adaptors/zensats/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const apyBase7d = growth(pps7d, 7);
+    const apyBase1d = growth(pps1d, 1);
+
+    return {
+      pool: `${v.address}-${CHAIN}`.toLowerCase(),
+      chain: utils.formatChain(CHAIN),
+      project: PROJECT,
+      symbol: v.symbol,
+      tvlUsd: (total / unit) * price,
+      apyBase: apyBase7d ?? apyBase1d ?? 0,
```
0xkr3p: remove apyBase1d and 0 fallbacks

src/adaptors/zensats/index.js
```diff
@@ -0,0 +1,136 @@
...
+      ...(apyBase7d !== null && { apyBase7d }),
```
0xkr3p: apyBase7d can be removed here as we are just repeating whats in apyBase

0xkr3p (commented):
hi @NateBrune, thanks for the PR - a couple of small tweaks before we can proceed

0xkr3p (approved):

PR #2859, opened 2026-07-30
Add Surge Credit lending yields adapter

punithbm:
Adds a yields adapter for Surge Credit, a BTC-collateralized USDC lending protocol on Base.

The protocol is already listed (DefiLlama-Adapters TVL PR is merged and live at https://defillama.com/protocol/surge-credit).

- Slug: `surge-credit`
- protocolId: `8328`
- Chain: Base (8453)
- Adapter: `src/adaptors/surge-credit/index.js` (single new file)

## Pools

One pool per on-chain market on the LiquidityPool contract `0xEE755F1BbcbF6e3260469D0f473522d71d3bdDda`:

- Market 0: variable rate (adaptive)
- Market 1: fixed rate

## How the numbers are computed

Everything is read live on-chain from the pool contract via `@defillama/sdk` (no external API or fetch, no hardcoded rates):

- `apyBase` (lender supply APY) = `borrowRate * utilization * (1 - reserveRate)`, then annualized as `e^r - 1` because the pool accrues interest with continuous compounding (`ExpMath.wCompoundFactor`).
- `apyBaseBorrow` = the current borrow rate from `getMarketBorrowRate`.
- `reserveRate` is read live per market from `markets(id).reserveRateBps` (currently 20% on both), so the supply APY stays correct if it changes.
- `totalSupplyUsd` and `totalBorrowUsd` come from the market struct.
- `tvlUsd` = available liquidity (supplied minus borrowed), from `getAvailableLiquidity`.
- `ltv` = `maxLtvBps`.

USDC is the only underlying, priced at ~$1.

## Expected low tvlUsd

The two markets currently run at 97 to 100 percent utilization, so available liquidity (`tvlUsd`) is small right now (market 0 about $7k, market 1 about $1). The test prints the standard "pools with <10k TVL won't be shown" notice. This is expected: `tvlUsd` is the correct lending value (supplied minus borrowed), not gross supply. The pools store fine and become visible once available liquidity rises above the threshold. `totalSupplyUsd` (about $261k and $108k) and the APYs are stored regardless.

## Testing

`npm run test --adapter=surge-credit` passes all 27 checks against live Base mainnet.

Only one file is added. No changes to `package.json`, lockfiles, or any other adapter. "Allow edits by maintainers" is enabled.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking Surge Credit lending markets on Base.
  * Displays supply and borrow APYs, utilization, TVL, available borrow, loan-to-value data, and market details.
  * Provides market-specific links and protocol information.
* **Bug Fixes**
  * Excludes inactive or unavailable markets from results.
  * Improved APY calculations for supply and borrowing.
  * Refined TVL and available-borrow calculations based on idle supplied liquidity.


src/adaptors/surge-credit/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+      const totalSupplyUsd = Number(market.totalSupplyAssets) / 1e6
+      const totalBorrowUsd = Number(market.totalBorrowAssets) / 1e6
+
+      pools.push({
+        pool: `surge-credit-${m}-${CHAIN}`,
+        chain: utils.formatChain(CHAIN),
+        project: 'surge-credit',
+        symbol: 'USDC',
+        tvlUsd: maxBorrowUsd,
```
0xkr3p: tvl formula should be totalSupplyUsd - totalBorrowUsd

src/adaptors/surge-credit/index.js
```diff
@@ -0,0 +1,93 @@
...
+      // reserveRateBps is read live per market (2000 = 20% on both today, not the
+      // 11%/14% in stale docs); it moves supply APY directly.
+      const reserveBps = Number(market.reserveRateBps)
+      const supplyAprBps = (borrowRateBps * utilBps * (1e4 - reserveBps)) / (1e4 * 1e4)
+
+      const totalSupplyUsd = Number(market.totalSupplyAssets) / 1e6
+      const totalBorrowUsd = Number(market.totalBorrowAssets) / 1e6
+
+      pools.push({
+        pool: `surge-credit-${m}-${CHAIN}`,
```
0xkr3p: better to use address here, such as `${LIQUIDITY_POOL.toLowerCase()}-${m}-${chain}`

0xkr3p (commented):
hey @punithbm, thanks for the PR a couple comments to resolve before we can proceed - thanks!

punithbm:
@0xkr3p  Thanks for the review - addressed both comments and fixed the TVL methodology.
1. tvlUsd was double-counting across the two markets. The previous version used getMaxBorrowAmount(marketId) as tvlUsd for both pools. That call returns the shared remaining borrow capacity, which is Market 0's idle liquidity, so it returns the identical value for both markets (currently $3,792.94 each) and the same USDC was counted twice.  

Fixed to the standard lending convention: tvlUsd = totalSupplyUsd - totalBorrowUsd (available/idle liquidity), computed per market from that market's own supply and borrow. The two markets' supplied figures are a disjoint partition of the same lender base. When a lender's capital is allocated to a fixed loan it moves out of Market 0's supply and into Market 1's, so each dollar is now counted once.  

2. Pool IDs now use the contract address as suggested: ${LIQUIDITY_POOL.toLowerCase()}-${marketId}-${chain}.  

3. On how we're representing the fixed market, want your input on the best convention. Market 1 (fixed rate) holds no idle liquidity of its own. It is funded on demand from Market 0 lenders' opt-in exposure and runs at ~100% utilization by construction. Right now we report it as:  
- tvlUsd: 0 and availableBorrowUsd: 0 (no independent idle liquidity) 
- totalSupplyUsd ($110,762) and totalBorrowUsd ($110,762) still reflect its real size  

So the fixed market appears with its true supplied/borrowed figures but a $0 headline TVL, which I think can be better represented given it does not hold separate capital. So is there a convention you would rather we follow?

0xkr3p:
hi @punithbm, thanks for the changes. The adapter APY figure seems slightly overstated compared to the surge UI  - is there any reason for this? Happy to proceed with TVL as is for fixed market for now

PR #2862, opened 2026-07-30
Add Neverland Isolated Market and Merkl reward APR

0xCatalyst:
Adds Neverland's Pendle AUSD isolated market and stops discarding Merkl reward APR.

### Isolated markets

Neverland runs self-contained markets alongside the canonical one, each with its own `PoolAddressesProvider`, `Pool` and `PoolDataProvider`. Only the canonical provider is registered in the shared `PoolAddressesProviderRegistry`, so isolated ones cannot be discovered onchain and are enumerated in a `markets` list; adding the next one is a single entry. Because collateral in one market cannot back debt in another, every reserve now carries `routeGroupKey` set to its data provider, matching `aave-v3`, so borrow routers do not pair reserves across markets.

### Merkl

`neverland` is on the merkl adapter's `protocolsBlacklist` ([`src/adaptors/merkl/index.js:22`](https://github.com/DefiLlama/yield-server/blob/18d9afd6783a7cc18c0094873c231e50b9882d4d/src/adaptors/merkl/index.js#L22)), and `addMerklRewardApy` only fills a reward field the adapter left empty. Every incentivised reserve already carries an onchain DUST emission, so the Merkl APR was reported in neither place.

### Reward expiry

`rewardApy` now skips streams past their `distributionEnd`. The controller does not zero `emissionPerSecond` when a distribution ends, so an expired campaign would otherwise keep paying indefinitely. Same guard as `gloop`, `zerolend` and `mahalend`. Inert today, since all 28 live streams end in 2030, but the value was already in the `getRewardsData` tuple so the check is free.

### Refactor

Adding the isolated market meant removing single-market assumptions throughout the adapter, so I restructured it instead of bolting on a second fetch path. It now follows three stages: fetch, build, then merge rewards, making the isolated market just another config entry.

That cleanup also collapsed duplicated multicalls and ABI lookups into helpers, replaced index-based reward tracking with a token-address map, removed two redundant RPC calls, switched to the shared `../aave-v3/poolAbi`, and dropped an unused `axios` import.

Two changes also fix correctness issues: reserve token addresses now come from `getReserveTokensAddresses` instead of assuming `getAllATokens()` ordering, and `rewardTokens` now includes borrow-side rewards, so debt-only incentives no longer report reward APY without listing the reward token.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lending market APY, TVL, borrow, liquidity, and cap calculations with better filtering of inactive/frozen reserves.
  * Refined reward APY calculations for DUST and Merkl, excluding expired reward emissions and improving reward token discovery.
  * Strengthened veDUST pool estimation using next-epoch revenue and more robust handling of missing/zero inputs.
  * Isolated errors so one market’s data issues no longer block the other’s results.


0xkr3p (approved):
hey @0xCatalyst,  thanks for the pr lgtm - approved

PR #2871, opened 2026-07-31
add sentora yield adapter

treeoflife2:
Adds a yields adapter for Sentora (protocolId 6807).

## What it lists

Reshapes Sentora's own vault API (services.vaults.sentora.com/vaults) into pools. Only the vaults that have no host adapter today, so nothing is duplicated:

- Upshift-tech Smart Vaults (Ethereum): Sentora USD, PRIME, ETH, BTC
- Kamino kvaults (Solana): Kamino Sentora PYUSD, Kamino USDG Ethena

Vaults hosted on Morpho (protocol morpho) and Euler (protocol eulerv2) are intentionally excluded, since morpho-blue and euler-v2 already own those pools.

TVL and apyBase/apyBase7d come from the API's per-vault analytics block. TVL is verifiable on-chain (upshift getTotalAssets, kamino api) and was spot-checked to match (e.g. sentUSD getTotalAssets = 40.6M = API tvlUsd). Dust vaults below 1000 USD are filtered.

Sourcing from the API means new vaults appear automatically once Sentora adds them, no code change needed.

## Open question

Whether Sentora is the right home for the Kamino/Upshift pools is left for @felix and @slasher125. A per-pool curator field would let host-owned pools (Morpho/Euler) surface on Sentora's page without duplicating data. See the adapter comment and PR discussion.

## Test

npm test --adapter=sentora: 41/41 pass.

### DIFF
--- a/src/adaptors/sentora/index.js
+++ b/src/adaptors/sentora/index.js
@@ -1,14 +1,21 @@
 const axios = require('axios');
 const utils = require('../utils');
 
-// Sentora's own canonical vault registry (Smart Vaults + curated Morpho/Euler/
-// Kamino vaults). Each vault carries an `analytics` block with tvlUsd and
-// apy7d/apy30d. TVL is verifiable on-chain (erc4626 totalAssets / upshift
+// Sentora's own canonical vault registry. Each vault carries an `analytics`
+// block with tvlUsd and apy7d/apy30d. TVL is verifiable on-chain (upshift
 // getTotalAssets / kamino api), spot-checked and matches. Reshaping this means
-// new vaults (e.g. the mWIN Morpho vault) appear automatically once Sentora
-// adds them to their API, no code change needed.
+// new vaults appear automatically once Sentora adds them to their API.
 const API = 'https://services.vaults.sentora.com/vaults';
 const MIN_TVL_USD = 1000; // drop dust/seed vaults
+
+// Host protocols already tracked by their own adapter (morpho-blue, euler-v2).
+// Listing their vaults here too would duplicate data, so we skip them and only
+// surface the genuinely-new pools (Kamino kvaults, Upshift-tech Smart Vaults)
+// that have no host adapter today.
+// Open question (Felix / slasher): a per-pool `curator` field would let a
+// morpho/euler pool show on Sentora's page without duplicating the row. Until
+// that exists, those pools stay under their host protocol only.
+const HOST_TRACKED = new Set(['morpho', 'eulerv2']);
 
 const apy = async () => {
   const { data } = await axios.get(API);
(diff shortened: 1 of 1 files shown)


src/adaptors/sentora/index.js
```diff
@@ -0,0 +1,51 @@
...
+  return vaults
+    .map((v) => {
+      const a = v.analytics;
+      const dt = v.depositToken;
+      const tvlUsd = Number(a?.tvlUsd);
+      if (v.status !== 'ACTIVE' || !dt?.address || !(tvlUsd >= MIN_TVL_USD))
+        return null;
+
+      const chain = utils.formatChain(v.blockchain?.name?.toLowerCase());
+      const apy7d = a.apy7d != null ? Number(a.apy7d) * 100 : null;
```
0xkr3p, line 38: i believe the apy data from their API is inclusive of rewards, reward apy should be a separate field `apyReward` along with `rewardTokens`

src/adaptors/sentora/index.js
```diff
@@ -0,0 +1,51 @@
...
+        pool: `${v.address}-${chain}`,
+        chain,
+        project: 'sentora',
+        symbol: dt.symbol,
+        poolMeta: v.name,
+        tvlUsd,
+        apyBase: apy30d ?? apy7d ?? 0,
+        ...(apy7d != null && { apyBase7d: apy7d }),
+        underlyingTokens: [dt.address],
+        url: v.landingUrl || 'https://vaults.sentora.com/',
```
0xkr3p, line 51: we could add `pricePerShare`

treeoflife2:
Good call on the duplication. Pushed 74914402 which drops the vaults hosted on Morpho and Euler (protocol morpho / eulerv2), since morpho-blue and euler-v2 are their proper owners. The adapter now lists only the 6 genuinely-new pools that have no host adapter today:

- Kamino kvaults (Solana): Kamino Sentora PYUSD, Kamino USDG Ethena
- Upshift-tech Smart Vaults (Ethereum): Sentora USD, PRIME, ETH, BTC

For reference, none of the 13 were actually double-listed in the pools API yet, but the principle stands so I removed the 7 host-owned ones anyway.

On the bigger question of whether Sentora is the right home for the Kamino/Upshift pools, I have left that open for @felix and @slasher125. The curator field is the clean answer: tag a pool with its curator and render the Yields tab on the curator's protocol page, so a Morpho or Euler pool can appear under both Morpho and Sentora without duplicating the row. That needs support in defillama-server and the frontend plus a new allowed field here, so happy to spec it out once you two agree on the shape. I left a note pointing at this in the adapter comment.

0xkr3p:
hi @treeoflife2, i've made the necessary changes in #2887

PR #2878, opened 2026-08-03
dinero-(pxeth): fall back to on-chain apxETH rate when their api returns NaN

0xshubhs:
dinero-(pxeth) has no pools in the yields api right now even though the protocol has ~$3-4m tvl and isnt in exclude.js.

cause is upstream: https://dinero.xyz/api/apr currently returns `{"apxEth":"NaN","pxEth":"3.30","sDinero":"7.04","apxSol":"9.63"}`. the adapter does `apyBase: Number(apyData.apxEth)`, so apyBase comes out NaN. triggerAdaptor nullifies non-finite apy fields, and since this adapter only sets apyBase, all three of apy/apyBase/apyReward end up null and the pool is dropped by the 'remove pools where all 3 apy related fields are null' filter. the whole protocol silently disappears from the product, no error anywhere.

apxETH is a standard erc4626 vault (asset() is pxETH 0x04c154b6...), so the rate is derivable on chain. added a fallback that only runs when the api value isnt finite: read convertToAssets(1e18) now and 30 days back and annualise. picked 30d because the vault only harvests every couple of weeks, pricePerShare has been flat at 1.132708080 for the last 7 days and last moved between d-14 (1.132162039) and d-7, so a 1d or 7d window reports 0%.

with the api value present nothing changes, it is still used as is. tested both ways: today (api NaN) gives apyBase 1.935%, which matches the 1.132708080/1.130924659 30d move annualised by hand, and the pool now survives the filters. if the api recovers, Number(apyData.apxEth) is finite and the fallback never runs.

if both sources are unusable it now throws instead of emitting NaN, so it shows up as a failing adapter rather than a protocol quietly vanishing.

separate thing i noticed but left alone: tvlUsd is totalSupply(shares) * ethPrice, but shares arent 1:1 with assets (pricePerShare is 1.133), so tvl reads ~12% low. happy to send that as its own PR.

## Summary by CodeRabbit

* **New Features**
  * Added on-chain APY calculation based on share-price performance over a 30-day period.
  * Re-enabled support for the pxETH protocol.

* **Bug Fixes**
  * Improved APY reliability by validating historical share prices and rejecting invalid or non-increasing values.
  * APY is no longer reported when a valid calculation cannot be completed.

### DIFF
--- a/src/adaptors/dinero-(pxeth)/index.js
+++ b/src/adaptors/dinero-(pxeth)/index.js
@@ -1,8 +1,44 @@
 const sdk = require('@defillama/sdk');
 const axios = require('axios');
-const { getPriceApiData } = require('../utils');
+const { getPriceApiData, getPriceApiUrl } = require('../utils');
 
 const token = '0x9ba021b0a9b958b5e75ce9f6dff97c7ee52cb3e6';
+const convertToAssetsAbi =
+  'function convertToAssets(uint256 shares) external view returns (uint256)';
+const SHARE_UNIT = '1000000000000000000';
+const LOOKBACK_DAYS = 30;
+
+const getBlock = async (timestamp) =>
+  (await axios.get(getPriceApiUrl(`/block/ethereum/${timestamp}`))).data.height;
+
+const pricePerShareAt = async (block) =>
+  Number(
+    (
+      await sdk.api.abi.call({
+        target: token,
+        block,
+        abi: convertToAssetsAbi,
+        params: [SHARE_UNIT],
+        chain: 'ethereum',
+      })
+    ).output
+  );
+
+const getOnChainApy = async () => {
+  const now = Math.floor(Date.now() / 1e3);
+  const [blockNow, blockThen] = await Promise.all([
+    getBlock(now),
+    getBlock(now - LOOKBACK_DAYS * 86400),
+  ]);
+  const [priceNow, priceThen] = await Promise.all([
+    pricePerShareAt(blockNow),
+    pricePerShareAt(blockThen),
+  ]);
+  if (!priceThen || !Number.isFinite(priceNow) || !Number.isFinite(priceThen))
+    return null;
+
+  return ((priceNow / priceThen) ** (365 / LOOKBACK_DAYS) - 1) * 100;
+};
 
 const getApy = async () => {
   const tvl =
@@ -12,6 +48,13 @@ const getApy = async () => {
   const priceKey = 'ethereum:0x0000000000000000000000000000000000000000';
   const ethPrice = (await getPriceApiData(`/prices/current/${priceKey}`)).coins[priceKey]?.price;
 
+  let apyBase = Number(apyData.apxEth);
+  if (!Number.isFinite(apyBase)) apyBase = await getOnChainApy();
+  if (!Number.isFinite(apyBase))
+    throw new Error(
+      'dinero: apxETH apy unavailable from dinero.xyz/api/apr and from on-chain pricePerShare'
+    );
+
   return [
     {
       pool: token,
@@ -19,7 +62,7 @@ const getApy = async () => {
       project: 'dinero-(pxeth)',
       symbol: 'apxeth',
       tvlUsd: tvl * ethPrice,
-      apyBase: Number(apyData.apxEth),
+      apyBase,
       underlyingTokens: ['0x0000000000000000000000000000000000000000'],
       searchTokenOverride: token, //autocompounding Pirex Ether
       isIntrinsicSource: true,
--- a/src/adaptors/dinero-(pxeth)/index.js
+++ b/src/adaptors/dinero-(pxeth)/index.js
@@ -6,7 +6,7 @@ const token = '0x9ba021b0a9b958b5e75ce9f6dff97c7ee52cb3e6';
 const convertToAssetsAbi =
   'function convertToAssets(uint256 shares) external view returns (uint256)';
 const SHARE_UNIT = '1000000000000000000';
-const LOOKBACK_DAYS = 30;
+const LOOKBACK_DAYS = 180;
 
 const getBlock = async (timestamp) =>
   (await axios.get(getPriceApiUrl(`/block/ethereum/${timestamp}`))).data.height;
@@ -40,7 +40,14 @@ const getOnChainApy = async () => {
     !Number.isFinite(priceNow) ||
     !Number.isFinite(priceThen)
   )
-    return null;
+    throw new Error(
+      `dinero: bad apxETH convertToAssets reads (${priceThen} -> ${priceNow})`
+    );
+
+  if (priceNow <= priceThen)
+    throw new Error(
+      `dinero: apxETH share price hasnt moved in ${LOOKBACK_DAYS}d (${priceThen} -> ${priceNow}), no harvest to measure`
+    );
 
   return ((priceNow / priceThen) ** (365 / LOOKBACK_DAYS) - 1) * 100;
 };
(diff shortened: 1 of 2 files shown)


0xshubhs:
took both, good catches.

**null/blank api value** — this was a real hole. `Number(null)` and `Number('')` are both 0, which is finite, so if dinero swapped "NaN" for null or an empty string the adapter would publish apyBase 0 and never reach the fallback. now it treats null/undefined/blank as missing and falls through to the on-chain rate. a genuine 0 from the api is still respected as 0, since that's a real value.

checked each shape: "NaN", null, '', '   ', undefined all fall back; '3.30' and a literal 0 still use the api value.

**zero priceNow** — also real. priceThen was guarded but priceNow wasn't, so a 0 read at the current block gives (0/x)^(365/30)-1 = -100%, and the negative clip downstream turns that into a published 0% rather than an error. both ends are checked now.

re-ran after the change, still 1.93%.

0xshubhs:
found something that changes this PR: dinero-(pxeth) is already in exclude.js (`{ id: '3912', slug: 'dinero-(pxeth)' }`, added 07-20 in adce86a4). i missed it at first because my checkout was behind.

so the adapter fix alone wouldn't have brought the pool back, same as amnis #2793 and yieldfi #2843 where the exclude entry had to come out separately once the source was working. pushed a second commit removing just that line.

reasoning for removing it now: the reason the adapter was producing nothing is that dinero.xyz/api/apr serves `"apxEth":"NaN"`, and with the fallback the adapter no longer depends on that field at all, so it returns a real number whether or not they fix it upstream. it also isn't a dead protocol, dinero.xyz and the deposit page are up, docs are up, the same endpoint still serves valid pxEth/sDinero/apxSol values, the dinero-protocol github org pushed yesterday, and tvl is ~$3.9m at -5% over 7d.

the entry has no comment on it, unlike the altura one right below it (`// deposits closed and sus activity`), so i'm assuming it was excluded for the broken data rather than anything about the protocol. if it was excluded for a reason i can't see from outside, say so and i'll drop that commit and leave the exclude in place.

adapter after both commits: 1 pool, apyBase 1.93%, tvlUsd $3.05m.

0xkr3p:
hi @0xshubhs, thanks for the changes, I think we can drop the api and just have the onchain resolution?

0xshubhs:
done, api call is gone and apy comes only from `convertToAssets` now.

one thing i want to flag before you merge, because dropping the api exposed it. i had the lookback at 30d, and on this vault that's not safe. harvests are lumpy, so share price sits perfectly flat for long stretches and a short window reads that as 0% rather than "no data".

pulled `convertToAssets(1e18)` every 6 days for the last 480d and simulated what each window would have published:

```
window   samples   readings under 0.01%   range            today
   1d      78          23/78              0.00% - 848.09%   4.46%
   7d      78          23/78              0.00% -  37.90%   0.62%
  30d      74          15/74              0.00% -  17.72%   2.22%
  60d      69           7/69              0.00% -   9.43%   9.43%
  90d      64           2/64              0.00% -   6.19%   6.19%
 180d      49           0/49              2.45% -   5.28%   4.87%
 365d      18           0/18              3.27% -   4.33%   4.26%
```

two flat stretches did it: 2025-12-15 -> 2026-01-26 (42d) and 2026-03-03 -> 2026-06-07 (96d). so 30d would have published a hard 0% on 15 of 74 sampled days, which is the same silent zero the pool was excluded for, just from a different direction.

moved it to 180d, the shortest window that never went to zero across the whole history. sanity check on the level: share price went 1.075370 -> 1.132708 between 2025-04-13 and 2026-07-31, 5.33% over 474d = **4.08% annualized realized**. 180d today gives 4.12% and 365d gives 4.26%, while 30d gives 2.22% and 60d gives 9.43% purely from where the harvests land. price is monotonic over all 78 steps, no decreases.

also added a guard that throws when price is unchanged across the window instead of returning 0, so if harvests ever stop for 180d the pool drops out rather than showing 0%.

adapter now: 1 pool, apyBase 4.12%, tvlUsd $3.04m, 13/13 local tests pass.

happy to use 90d or 365d instead if you'd rather, 180 was just the shortest one with no zeros in it.

separately, the red `test` check on this pr isn't the adapter, it's the ci escaping parens in the dir name:

```
Adapter not found. Tried:
 - /home/runner/work/yield-server/yield-server/src/adaptors/dinero-\\(pxeth\\)/index.js
```

`changed-files` has `safe_output: true` by default so it backslash escapes the path, and the name reaches `npm run test --adapter=` unquoted. hits all three parenthesised adapters (`dinero-(pxeth)`, `bfx-(blast-futures)`, `genesislrt-(native-restaking)`) so their tests can never go green. locally `npm run test --adapter='dinero-(pxeth)'` passes 13/13. can send that as its own pr if you want it fixed, ser

0xkr3p:
> done, api call is gone and apy comes only from `convertToAssets` now.
> 
> one thing i want to flag before you merge, because dropping the api exposed it. i had the lookback at 30d, and on this vault that's not safe. harvests are lumpy, so share price sits perfectly flat for long stretches and a short window reads that as 0% rather than "no data".
> 
> pulled `convertToAssets(1e18)` every 6 days for the last 480d and simulated what each window would have published:
> 
> ```
> window   samples   readings under 0.01%   range            today
>    1d      78          23/78              0.00% - 848.09%   4.46%
>    7d      78          23/78              0.00% -  37.90%   0.62%
>   30d      74          15/74              0.00% -  17.72%   2.22%
>   60d      69           7/69              0.00% -   9.43%   9.43%
>   90d      64           2/64              0.00% -   6.19%   6.19%
>  180d      49           0/49              2.45% -   5.28%   4.87%
>  365d      18           0/18              3.27% -   4.33%   4.26%
> ```
> 
> two flat stretches did it: 2025-12-15 -> 2026-01-26 (42d) and 2026-03-03 -> 2026-06-07 (96d). so 30d would have published a hard 0% on 15 of 74 sampled days, which is the same silent zero the pool was excluded for, just from a different direction.
> 
> moved it to 180d, the shortest window that never went to zero across the whole history. sanity check on the level: share price went 1.075370 -> 1.132708 between 2025-04-13 and 2026-07-31, 5.33% over 474d = **4.08% annualized realized**. 180d today gives 4.12% and 365d gives 4.26%, while 30d gives 2.22% and 60d gives 9.43% purely from where the harvests land. price is monotonic over all 78 steps, no decreases.
> 
> also added a guard that throws when price is unchanged across the window instead of returning 0, so if harvests ever stop for 180d the pool drops out rather than showing 0%.
> 
> adapter now: 1 pool, apyBase 4.12%, tvlUsd $3.04m, 13/13 local tests pass.
> 
> happy to use 90d or 365d instead if you'd rather, 180 was just the shortest one with no zeros in it.
> 
> separately, the red `test` check on this pr isn't the adapter, it's the ci escaping parens in the dir name:
> 
> ```
> Adapter not found. Tried:
>  - /home/runner/work/yield-server/yield-server/src/adaptors/dinero-\\(pxeth\\)/index.js
> ```
> 
> `changed-files` has `safe_output: true` by default so it backslash escapes the path, and the name reaches `npm run test --adapter=` unquoted. hits all three parenthesised adapters (`dinero-(pxeth)`, `bfx-(blast-futures)`, `genesislrt-(native-restaking)`) so their tests can never go green. locally `npm run test --adapter='dinero-(pxeth)'` passes 13/13. can send that as its own pr if you want it fixed, ser

hi @0xshubhs,  thank you for the thorough investigation - i think the period of 180d is too large of a window, we typically do not use a lookback of more than 30d simply because they're too slow to display recent changes. I would keep at 30d, then we can keep an eye on the stability

0xshubhs:
done, back to 30d.

that works out fine because of the guard, and i checked rather than assuming. re-ran the whole 480d history at 30d and split it by what the adapter would actually do:

```
samples       79
publishes     64   apy range 0.30% - 17.68%
guard throws  15
```

the 15 are exactly the days i was worried about. price is bit-for-bit unchanged across the window there, so `priceNow <= priceThen` fires and the pool drops out for that run instead of publishing a 0%:

```
2026-04-01  then=1116059747913468300  now=1116059747913468300  -> throws
2026-03-27  then=1115811293552730500  now=1116059747913468300  -> publishes 0.27%
```

so no zero ever reaches the api at 30d, the pool just goes missing on those days. that's the tradeoff you're accepting and it's the safe direction. all 15 fall in the two harvest gaps (2026-01-13..01-25 and 2026-04-01..06-06), so if harvests keep their current cadence it won't come up.

adapter now: 1 pool, apyBase 1.75%, tvlUsd $3.08m, 13/13 local tests pass. it reads lower than the 4.08% realized over the last 474d, which is the recency you wanted.

the red `test` check is still the ci paren escaping, not this adapter, ser

0xshubhs:
also took coderabbit's open one on tvl, it was right and it's a bigger error than the apy question.

`totalSupply` is apxETH shares, and the adapter was multiplying that straight by the eth price as if a share were an eth. it isn't, that's the same 1.13 exchange rate the apy is derived from. so tvl was understated by the whole accrued yield:

```
before   $3,075,658     totalSupply * ethPrice
after    $3,476,798     convertToAssets(totalSupply) * ethPrice
ratio    1.1304         = current share price
```

reused the existing `convertToAssets` call for it rather than adding a second path, `pricePerShareAt` is now just that with a 1e18 argument.

13/13 still pass, ser

0xkr3p (approved):
approved nice job @0xshubhs

PR #2884, opened 2026-08-05
liminal-basis: split tvl across chains by share supply

### LINKED ISSUE #2869: liminal-basis: one vault NAV emitted on 3 chains, ~$46.5M of duplicated pool TVL
`src/adaptors/liminal-basis/index.js` reads a single global NAV from the Hyperliquid L1 hub oracle and assigns that same `tvlUsd` to every deployment of the product:

```js
const tvlUsd = await getTvlUsd(hubDeployment.navOracle)
...
Object.entries(product.deployments).map(... tvlUsd ...)
```

So hyperliquid, arbitrum and ethereum each report the full NAV.

## Live right now

`yields.llama.fi/pools`, 12 pools:

```
Hyperliquid L1 LIMUSD  10,875,962     Ethereum LIMUSD  10,875,962     Arbitrum LIMUSD  10,875,962
Hyperliquid L1 XHYPE    5,772,022     Ethereum XHYPE    5,772,022     Arbitrum XHYPE    5,772,022
Hyperliquid L1 XLEND    5,387,872     Ethereum XLEND    5,387,872     Arbitrum XLEND    5,387,872
Hyperliquid L1 XBTC     1,227,327     Ethereum XBTC     1,227,327     Arbitrum XBTC     1,227,327

sum reported: 69,789,549   real NAV: ~23,262,000
```

For comparison the TVL side reports `currentChainTvls: {"Hyperliquid L1": 25321673}` and nothing on Arbitrum or Ethereum.

## The spokes are tiny

On-chain `totalSupply`:

| | hub (HyperEVM) | arbitrum | ethereum |
|---|---|---|---|
| limUSD | 10,680,497.5 | 316,673.7 | 206.97 |
| xHYPE | 5,536,731.0 | 25,987.7 | 103.21 |

## The formula is already in the adapter

`apy.js:111` `calculateShareValue` computes share value as `NAV / hubTotalSupply`, reading supply from the hub `shareManagerAddress`. That is, the adapter already treats hub supply as the full share count, which is the lock-and-mint reading where spoke supplies are a subset of it.

Applying the adapter's own share value per chain:

```
limUSD arbitrum  316,673.7 x (NAV/hubSupply) ~= $322.5k   (reported: $10.88M)
limUSD ethereum      206.97 x (NAV/hubSupply) ~= $211      (reported: $10.88M)
limUSD hyperliquid                            ~= $10.55M
```

Those three sum back to NAV exactly.

## Scope

The 3x pattern predates #2838, it was already there for xHYPE. #2838 generalised it from one product to four, which took the duplicated total from roughly $11.5M to roughly $46.5M.

## Suggested fix

Scale `tvlUsd` per deployment by that chain's share supply, with the hub taking `hubSupply - sum(spokes)`. Emitting only the hub pool would also work but loses the per-chain breakdown.

@shkkgit since you wrote the adapter, does the lock-and-mint reading hold, i.e. is every spoke share backed by a hub share that is included in `hubTotalSupply`? If it is burn-and-mint instead the split needs a different denominator and I would rather have that confirmed than guess at it.

0xshubhs:
Fixes #2869. Diagnosis and the suggested approach are @BhariGowda's, i confirmed the open question on-chain and implemented it.

`getPools` read one NAV from the hub oracle and handed the same `tvlUsd` to all three deployments, so each product was counted three times. 12 pools, ~$69.7M reported against ~$23.2M of real NAV.

## the open question was lock-and-mint vs burn-and-mint

That was the blocker on the issue, and it decides the denominator, so i checked rather than assumed. Each spoke is a LayerZero OFT whose `peers(30367)` points at a **separate** HyperEVM contract, and that contract is a lockbox adapter holding the hub share token:

```
limUSD  peer 0x95bb9d9d46b80b654ac93bb6d20fcc557d27bb49
        token()            0x1822bd...83c8   == hub share token
        approvalRequired() true              (adapter, not a mint/burn oft)
        hub tokens locked  316,880.67
        arb + eth supply   316,880.67        <- equal to the wei
```

Same on all four:

```
          locked in adapter      arb+eth supply
limUSD        316,880.67           316,880.67
xHYPE          26,090.88            26,090.88
xBTC            1,628.76             1,628.76
xLEND               4.16                 4.16
```

So it is lock-and-mint: every spoke share is backed by a hub share that is still inside `hubTotalSupply`. The lock-and-mint reading in the issue holds, and `hubSupply - sum(spokes)` is the right hub figure.

Independent check on the same conclusion, `NAV / (hub + spokes)` gives limUSD a share value of 0.990007, i.e. below par for a vault that only accrues, while `NAV / hubSupply` gives 1.019537, in line with its siblings (1.0175 - 1.0440).

## fix

Read `totalSupply` (and `decimals`) per deployment, value every chain at the hub share value `NAV / hubSupply`, and give the hub `hubSupply - sum(spokes)`.

## result

Each product now sums back to its NAV exactly:

```
xHYPE   published $ 5,821,405   NAV $ 5,821,405   diff -0.00
xBTC    published $ 1,228,707   NAV $ 1,228,707   diff -0.00
xLEND   published $ 5,334,525   NAV $ 5,334,525   diff -0.00
limUSD  published $10,828,731   NAV $10,828,731   diff  0.00

TOTAL   published $23,213,368   NAV $23,213,368   diff  $0.00
```

was $69.7M, so this drops ~$46.4M of duplicated pool TVL while keeping the per-chain breakdown.

72/72 tests pass. Pool count goes 12 -> 11: xLEND on ethereum has zero share supply, so it values to 0 and the existing `tvlUsd > 0` filter drops it instead of showing a $5.3M phantom. apy is untouched, it was already a hub-level number.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pool TVL reporting by calculating values separately for each deployment.
  * Correctly accounts for bridged token supply and assigns remaining supply to the Hyperliquid hub.
  * Normalized token supply values for more consistent and accurate displays.
  * Pools are no longer shown when the hub has no available supply or bridged supply is excessive.
  * Improved consistency and accuracy of TVL and supply data across supported products and deployments.
  * Improved data retrieval efficiency for more timely pool information.

### DIFF
--- a/src/adaptors/liminal-basis/index.js
+++ b/src/adaptors/liminal-basis/index.js
@@ -167,6 +167,8 @@ async function getPools(product, nav, supplies) {
       key === 'hyperliquidL1' ? total : total.plus(supplies[key]),
     new BigNumber(0)
   );
+  if (bridgedSupply.gt(hubSupply)) return [];
+
   const shareValue = new BigNumber(nav).div(hubSupply);
 
   return deployments.map(([key, config]) => {
--- a/src/adaptors/liminal-basis/index.js
+++ b/src/adaptors/liminal-basis/index.js
@@ -112,27 +112,59 @@ async function getTvlUsd(navOracleAddress) {
   return navIn18Decimals.div(1e18).toNumber();
 }
 
+async function getShareSupply(address, chain) {
+  const [supplyResult, decimalsResult] = await Promise.all([
+    sdk.api.abi.call({ target: address, abi: abi.totalSupply, chain }),
+    sdk.api.abi.call({ target: address, abi: abi.decimals, chain }),
+  ]);
+
+  return new BigNumber(supplyResult.output).div(
+    new BigNumber(10).pow(decimalsResult.output)
+  );
+}
+
 async function getPools(product) {
   const hubDeployment = product.deployments.hyperliquidL1;
-  const tvlUsd = await getTvlUsd(hubDeployment.navOracle);
+  const nav = await getTvlUsd(hubDeployment.navOracle);
   const apyBase7d = await getApy7d(
     hubDeployment.navOracle,
     hubDeployment.address,
     HYPERLIQUID_L1_CHAIN
   );
 
-  return Object.entries(product.deployments).map(([key, config]) => ({
-    pool: `${config.address}-${config.chain}-${key}`.toLowerCase(),
-    chain: utils.formatChain(config.chain),
-    project: 'liminal-basis',
-    symbol: product.symbol,
-    tvlUsd,
-    apyBase: apyBase7d,
-    apyBase7d,
-    underlyingTokens: config.underlyingTokens,
-    poolMeta: product.poolMeta,
-    url: 'https://liminal.money/app/tokenized',
-  }));
+  const deployments = Object.entries(product.deployments);
+  const supplies = await Promise.all(
+    deployments.map(([, config]) => getShareSupply(config.address, config.chain))
+  );
+
+  const hubIndex = deployments.findIndex(([key]) => key === 'hyperliquidL1');
+  const hubSupply = supplies[hubIndex];
+  if (hubSupply.isZero()) return [];
+
+  const bridgedSupply = supplies.reduce(
+    (total, supply, index) =>
+      index === hubIndex ? total : total.plus(supply),
+    new BigNumber(0)
+  );
+  const shareValue = new BigNumber(nav).div(hubSupply);
+
+  return deployments.map(([key, config], index) => {
+    const shares =
+      index === hubIndex ? hubSupply.minus(bridgedSupply) : supplies[index];
+
+    return {
+      pool: `${config.address}-${config.chain}-${key}`.toLowerCase(),
+      chain: utils.formatChain(config.chain),
+      project: 'liminal-basis',
+      symbol: product.symbol,
+      tvlUsd: shares.times(shareValue).toNumber(),
+      apyBase: apyBase7d,
+      apyBase7d,
+      underlyingTokens: config.underlyingTokens,
+      poolMeta: product.poolMeta,
+      url: 'https://liminal.money/app/tokenized',
+    };
+  });
 }
 
 async function main() {
--- a/src/adaptors/liminal-basis/index.js
+++ b/src/adaptors/liminal-basis/index.js
@@ -101,56 +101,77 @@ const PRODUCTS = [
   },
 ];
 
-async function getTvlUsd(navOracleAddress) {
-  const navResult = await sdk.api.abi.call({
-    target: navOracleAddress,
+async function getTvlUsd(products) {
+  const navResults = await sdk.api.abi.multiCall({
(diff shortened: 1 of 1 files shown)


0xkr3p (commented):
hey @0xshubhs, changes look good, one minor improvement might be to optimize by using multicall to reduce the amount of rpc reqs

0xshubhs:
def , lemme try optimize it a bit

0xshubhs:
done, pushed.

the nav reads and the supply/decimals reads are batched now. supplies get grouped by chain across all 4 products, so instead of 12 deployments x 2 calls each it's one totalSupply multicall + one decimals multicall per chain, and the 4 getNAV oracle reads collapse into a single multicall on hyperliquid.

counted the actual json-rpc traffic on the wire before and after:

| | eth_call | pools | total tvl |
|---|---|---|---|
| before | 47 | 11 | $23,216,975.43 |
| after | 24 | 11 | $23,216,975.43 |

output is byte identical across three runs, same pools and same tvl to the cent.

the 24 that are left are almost all in apy.js — the 7d apy reads totalSupply/decimals at the two NAV-update blocks per product, and those blocks are different for each product so a multicall can't span them. i could still drop 8 of those by reading decimals once instead of at both historical blocks (it's the same share token, it isn't going to change), but that swaps a historical read for a current one so i left it alone. say the word and i'll do that too ser.

0xshubhs:
taken, guard is in.

on the mechanism though, the race as described can't actually produce the inversion. this is a lock-and-mint bridge — a hub->spoke transfer locks the hub shares in the adapter and mints on the spoke, so hub `totalSupply` doesn't move when something bridges. that's also why the split reconciles to nav exactly: `hubSupply` already counts the bridged shares, and hyperliquid-resident is the remainder. a spoke mint landing after the hyperliquid read can't push `bridgedSupply` past `hubSupply`, and both in-flight directions skew the safe way (spoke burns before the hub unlocks, adapter locks before the spoke mints).

current headroom, read just now:

| product | hub | bridged | bridged/hub |
|---|---|---|---|
| xHYPE | 5,576,215 | 26,090 | 0.47% |
| xBTC | 1,207,557 | 1,628 | 0.13% |
| xLEND | 5,223,830 | 4.16 | 0.0001% |
| limUSD | 10,622,998 | 316,880 | 2.98% |

but the consequence you describe is real if the invariant ever does break, so it's worth failing closed. i stubbed a spoke supply high to force it and ran it both ways:

- without the guard: 7 pools, $4,102,555,888,571,699 total, and **zero** negative pools — the negative hub entries get silently eaten by the `tvlUsd > 0` filter at the end, so the spokes just over-report and nothing looks wrong
- with the guard: 0 pools, nothing published

so it turns a silent over-report into an omission, which is the right way round for this adapter. skipping the product rather than retrying, since the reads are on unrelated chains and there's no common block to pin them to.

normal path is unchanged, 11 pools and $23,216,975.43 identical across runs.

0xkr3p (approved):
great job @0xshubhs thanks!

PR #2886, opened 2026-08-06
add neutral-trade adapter

jakeyvee:
Adds a yields adapter for Neutral Trade Solana bundle vaults running quantitative strategies on USDC deposits. Vault NAV is pushed onchain daily by a keeper oracle. Shares are virtual u128 values on the vault account with no Token share mint, so `pool` is the bundle account address and `token` is explicitly `null`.

Protocol slug `neutral-trade` (id `5548`) is listed on the DefiLlama TVL page.

### Pools

The adapter maps all 12 public, enabled, catalog-visible Solana USDC bundle vaults in the bundle-indexer registry, including Neutral Autopilot (legacy vault id `76`). During validation, the aggregate yields endpoint still returned its previous 11-pool set while that visibility change propagates; all 11 returned pools were above the $10k display floor. Autopilot's live daily snapshot is already available and its adapter row was validated separately.

### Data sources

- `https://api.neutral.trade/public/yields` provides a 60-second cached projection of each visible vault's TVL and after-fee 7-day APY. TVL combines the onchain token balance with keeper-oracle-posted equity deployed at external venues, matching the existing DefiLlama TVL adapter's definition.
- `https://www.neutral.trade/api/daily-vault-snapshot/<vaultId>?limit=1&sort=desc` provides each vault's latest token-denominated daily share-price close. The adapter also maps every bundle account to its specific strategy page.

These public read-only APIs are used because share-price history lives in program event logs. Reconstructing the same history from a public RPC requires a paced walk through signatures and transactions for every vault.

### APY methodology

`apyBase` and `apyBase7d` both use the same trailing 7-day share-price growth, annualized as `(pps_now / pps_then)^(365 / window) - 1` and net of performance and management fees. A vault without a complete 7-day observation is omitted until that window is available. A 24-hour window is not used because NAV updates once per day, so it can straddle one or two oracle updates. The protocol's points program is excluded, so no `apyReward` is emitted.

`pricePerShare` is the latest underlying-token amount per virtual share.

`npm run test --adapter=neutral-trade`: 94 passed against the current 11-pool live response. The Autopilot fixture-path check also passed with its live token PPS and pool-specific URL.

## Summary by CodeRabbit

* **New Features**
  * Added support for Neutral Trade vaults.
  * Added APY and price-per-share data retrieval with validation and retry handling.
  * Added standardized Solana pool records for supported vaults.
  * Converted APY values to percentage format and excluded invalid results.

### DIFF
--- a/src/adaptors/neutral-trade/index.js
+++ b/src/adaptors/neutral-trade/index.js
@@ -1,0 +1,41 @@
+const utils = require('../utils');
+
+// Public, unauthenticated endpoint served by Neutral Trade's bundle indexer.
+// Contract: { data: { vaults: [{ bundleKey, name, asset: { mint, symbol },
+// tvlUsd, apy7dAfterFees, apy30dAfterFees, apyInceptionAfterFees }] } }
+// with rates as decimal fractions net of performance + management fees.
+const API_URL = 'https://bundle-indexer-api.onrender.com/public/yields';
+
+const toPct = (rate) => (Number.isFinite(rate) ? rate * 100 : null);
+
+const getApy = async () => {
+  const { data } = await utils.withRetry(() => utils.getData(API_URL));
+
+  return data.vaults
+    .filter((vault) => vault.asset?.mint && vault.asset?.symbol)
+    .map((vault) => ({
+      pool: vault.bundleKey,
+      chain: utils.formatChain('solana'),
+      project: 'neutral-trade',
+      symbol: utils.formatSymbol(vault.asset.symbol),
+      poolMeta: vault.name ?? undefined,
+      underlyingTokens: [vault.asset.mint],
+      // null when the indexer has no fresh USD price; dropped by keepFinite
+      tvlUsd: vault.tvlUsd,
+      // 7d NAV growth net of fees; longer windows cover young vaults
+      apyBase: toPct(
+        vault.apy7dAfterFees ??
+          vault.apy30dAfterFees ??
+          vault.apyInceptionAfterFees
+      ),
+      apyBase7d: toPct(vault.apy7dAfterFees),
+    }))
+    .filter(utils.keepFinite);
+};
+
+module.exports = {
+  protocolId: '5548',
+  timetravel: false,
+  apy: getApy,
+  url: 'https://www.neutral.trade/',
+};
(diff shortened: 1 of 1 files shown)


src/adaptors/neutral-trade/index.js
```diff
@@ -0,0 +1,44 @@
...
+      symbol: utils.formatSymbol(vault.asset.symbol),
+      poolMeta: vault.name ?? undefined,
+      underlyingTokens: [vault.asset.mint],
+      // bundle shares are a virtual u128 on the vault account; no SPL share
+      // mint exists
+      token: null,
+      // null when the indexer has no fresh USD price; dropped by keepFinite
+      tvlUsd: vault.tvlUsd,
+      // 7d NAV growth net of fees; longer windows cover young vaults
+      apyBase: toPct(
```
0xkr3p: apyBase should be one value, we shouldn't mix the durations - recommend to stick with 7 if thats the most the stable, no higher than 30d though pls as this struggles to resemble recent perf

src/adaptors/neutral-trade/index.js
```diff
@@ -0,0 +1,44 @@
...
+        vault.apy7dAfterFees ??
+          vault.apy30dAfterFees ??
+          vault.apyInceptionAfterFees
+      ),
+      apyBase7d: toPct(vault.apy7dAfterFees),
```
0xkr3p, line 97: add `pricePerShare` and `url` with a pool specific url

0xkr3p (commented):
hi @jakeyvee, thanks for the PR - I've left a couple of small comments to resolve before we can move ahead

src/adaptors/neutral-trade/index.js
```diff
@@ -0,0 +1,44 @@
...
+const getApy = async () => {
+  const { data } = await utils.withRetry(() => utils.getData(API_URL));
+
+  return data.vaults
+    .filter((vault) => vault.asset?.mint && vault.asset?.symbol)
+    .map((vault) => ({
+      pool: vault.bundleKey,
+      chain: utils.formatChain('solana'),
+      project: 'neutral-trade',
+      symbol: utils.formatSymbol(vault.asset.symbol),
```
0xkr3p: remove format symbol func

src/adaptors/neutral-trade/index.js
jakeyvee: Resolved in 0d5adb82: apyBase and apyBase7d now both use only apy7dAfterFees; the duration fallback was removed.

src/adaptors/neutral-trade/index.js
jakeyvee: Resolved in 0d5adb82: added token-denominated pricePerShare from the latest daily snapshot and a vault-specific strategy URL for all 11 pools.

src/adaptors/neutral-trade/index.js
jakeyvee: Resolved in 0d5adb82: symbol now uses vault.asset.symbol directly.

0xkr3p (approved):

PR #2885, opened 2026-08-05
Origin adapters tvl apy

apexearth:
origin-arm listed only the Lido ARM. It now lists all five live ARMs (~$8.8M):
TVL from totalAssets(), APY from convertToAssets() over a trailing 30d window,
and Merkl incentives via the shared merkl helper as apyReward.

The OTokens valued TVL with vault.totalValue(), which counts the OToken their
AMOs mint into Curve/Aerodrome pools -- protocol liquidity, not deposits. That
roughly doubled the reported pool size (superOETHb $28M against ~$19M). They now
read Origin's protocolDailyStatDetails.tvl, which is already net of it, and
measure APY on-chain from rebasingCreditsPerTokenHighres over 7d.

superOETHb's pool TVL intentionally differs from Origin Ether's protocol TVL on
DefiLlama, which additionally nets out the bridged wOETH already counted on
Ethereum: right for protocol TVL, wrong for pool size, since that wOETH is what
backs superOETHb holders' balances.

superOETHp is dropped (retired). Smaller fixes: apyBase rather than apy, OUSD's
underlyingTokens (USDT/DAI -> USDC), and chain names.

also update url links for each of the products

## Summary by CodeRabbit

* **New Features**
  * Added dynamic discovery and tracking for Origin ARM vaults on Ethereum and Sonic.
  * Added support for OETH, superOETHb, OUSD, and Origin Sonic products.
  * Added Merkl incentive APY where available.

* **Bug Fixes**
  * Improved TVL and APY accuracy using Origin production statistics and on-chain yield data.
  * Excluded retired, unsupported, or invalid vault results.
  * Updated product links, metadata, pricing, and displayed underlying assets.
  * Removed unsupported superOETHp data from Plume.

PR #2890, opened 2026-08-06
origin-ether: fix TVL, measure APY on-chain, drop superOETHp

apexearth:

------

TVL came from vault.totalValue(), which counts the OToken the vault's own AMO strategies mint into Curve/Aerodrome pools -- protocol-owned liquidity, not depositor funds. That put superOETHb at ~$29M against ~$19M of holder-owned backing. It now reads Origin's protocolDailyStatDetails.tvl, which is already net of it.

That figure is deliberately not Origin Ether's protocol TVL on DefiLlama, which on Base additionally drops the bridged wOETH strategy because that backing is already counted on Ethereum. Netting it out is right for protocol TVL and wrong for pool size: the bridged wOETH is precisely what backs superOETHb holders' balances.

APY is now measured on-chain from rebasingCreditsPerTokenHighres over 7d, in a helper shared with the other origin-* adaptors. The squid's apy7DayAvg reads ~0.2pp low because it includes the in-progress day, annualised over a full day regardless of how much of it has elapsed.

superOETHp is dropped -- the Plume product is retired. Smaller fixes: apyBase rather than apy, a per-pool url (there was none), and hardcoded chain names instead of capitalizing the sdk chain key.

## Summary by CodeRabbit

* **New Features**
  * Added support for OETH on Ethereum and superOETHb on Base.
  * Added TVL reporting based on Origin protocol statistics and current ETH pricing.
  * Added on-chain base APY calculations using realized seven-day yield.
  * Pool details now identify native ETH as the underlying asset and include base APY information.
* **Bug Fixes**
  * Invalid, unavailable, or failed TVL and APY results are excluded from displayed data.
  * Improved token search labeling and pool metadata accuracy.

### DIFF
--- a/src/adaptors/origin-ether/index.js
+++ b/src/adaptors/origin-ether/index.js
@@ -1,125 +1,105 @@
-const axios = require('axios');
+/*
+ * Origin Ether: OETH (Ethereum) and superOETHb (Base).
+ *
+ * Both TVL and APY come from Origin's production squid, so these pools carry the same numbers
+ * Origin reports itself.
+ *
+ * `protocolDailyStatDetails.tvl` is the backing that belongs to holders: it nets out the OToken
+ * the vault's own AMO strategies minted into Curve/Aerodrome pools, which is protocol-owned
+ * liquidity rather than depositor funds. Raw `vault.totalValue()` includes it and puts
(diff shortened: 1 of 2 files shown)


0xkr3p (approved):
thanks for the PR @apexearth, approved

PR #2891, opened 2026-08-06
origin-arm: list every ARM, with on-chain TVL and APY

apexearth:

------

The adaptor listed only the Lido ARM. It now enumerates all five live ARMs (~$8.8M) from Origin's squid -- there is no on-chain registry to discover them from -- and reads every number on-chain or from Merkl.

TVL is totalAssets(), the net value backing the shares: it nets out the pending withdrawal queue and includes assets parked in the ARM's lending market, so it is the complete pool size. The previous balanceOf sum missed both.

APY is the growth of convertToAssets(1e18) over a trailing 30d window, replacing Origin's per-ARM REST endpoint, which only ever covered the Lido ARM. 30d rather than something shorter because ARM yield arrives in lumps: over 14d the OS ARM reads 0% and the Ethena ARM swings by 2pp. ARMs younger than the window, or recently reseeded, fall back to 14/7/3d. A share-price move above 0.5%/day is treated as a seeding event rather than yield -- the USDC ARM's price doubled overnight when it was funded out of its dust seed, which annualises to 7.1e9%.

Merkl incentives now come through the shared merkl helper as apyReward, and the figure is reported as apyBase rather than apy. A /block response without a height throws instead of silently falling back to the latest block, which would make both ends of the window the same sample and report exactly 0%. The wound down OS ARM is excluded.

## Summary by CodeRabbit

* **New Features**
  * Added automatic discovery of Origin ARM pools on Ethereum and Sonic.
  * Added pool details including TVL, base APY, share price, underlying assets, token addresses, and direct links.
  * Added Merkl reward APY information to supported pools.

* **Bug Fixes**
  * Excluded the discontinued Sonic ARM.
  * Excluded pools with missing data, invalid prices, unavailable history, or excessive share-price movements.
  * Improved resilience so failures on one network do not block data from others.

### DIFF
--- a/src/adaptors/origin-arm/index.js
+++ b/src/adaptors/origin-arm/index.js
@@ -1,56 +1,206 @@
+/*
+ * Origin ARM (Automated Redemption Manager) vaults.
+ *
+ * Every number here is read on-chain or from Merkl; Origin's squid is used only to enumerate the
+ * ARMs, since there is no on-chain registry to discover them from.
+ *
+ * Each ARM is a share vault denominated in its liquidity asset (`assets[0]`):
+ *   - `totalAssets()` is the net value backing the shares: it nets out the pending withdrawal
+ *     queue and includes assets parked in the ARM's lending market, so it is the complete pool TVL.
+ *   - `convertToAssets(1e18)` is the share price. Its trailing growth is the base yield.
+ *   - Merkl incentives come from the shared helper, which reads Merkl's own API.
+ */
+const { gql, request } = require('graphql-request');
 const sdk = require('@defillama/sdk');
+
 const utils = require('../utils');
+const { addMerklRewardApy } = require('../merkl/merkl-additional-reward');
 
-const WETH_ADDRESS = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2';
-const STETH_ADDRESS = '0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84';
-const ARM_WETH_STETH_ADDRESS = '0x85b78aca6deae198fbf201c82daf6ca21942acc6';
+const GRAPH_URL = 'https://origin.squids.live/origin-squid/graphql';
 
-const apy = async () => {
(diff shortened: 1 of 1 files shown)


0xkr3p (commented):
hi @apexearth, thanks for the PR - just small tweak before we can merge

src/adaptors/origin-arm/index.js
```diff
@@ -1,56 +1,206 @@
...
+        // Bare lowercase address: the format the Lido ARM has always used, kept so its
+        // existing yield history stays attached. Also what Merkl keys its opportunities on.
+        pool: address,
+        chain,
+        project: 'origin-arm',
+        symbol: arm.symbol,
+        tvlUsd: (Number(totalAssets[i]) / 10 ** arm.assetDecimals[0]) * price,
+        apyBase: apyBase[i],
+        underlyingTokens: arm.assets,
+        token: address,
```
0xkr3p, line 180: pls add `pricePerShare`

src/adaptors/origin-arm/index.js
```diff
@@ -1,56 +1,206 @@
...
+// An ARM earns single-digit percent a year, i.e. well under 0.05%/day. A larger move across the
+// window is a seeding or re-initialisation event, not yield -- the USDC ARM's share price doubled
+// overnight when it was funded out of its dust seed, which annualises to 7.1e9%.
+const MAX_DAILY_MOVE = 0.005;
+
+// Shorter windows tried, longest first, only for ARMs whose full window is unusable: either they
+// did not exist 30 days ago, or a jump like the above sits inside it. Everything else costs two
+// calls and never touches these. Kept short deliberately -- each entry is a round trip, and this
+// only applies to ARMs younger than the window or recently reseeded, which both age out of it.
+const FALLBACK_DAYS = [14, 7, 3];
```
0xkr3p: rather than using fallbacks, I suggest just setting `apyBase` as 7d window. Using a singular time window keeps our data consistent

apexearth:
Pushed changes for your feedback @0xkr3p

0xkr3p (approved):
sweet - approved @apexearth

PR #2893, opened 2026-08-06
origin-sonic: fix TVL, measure APY on-chain, fix the pool url

apexearth:

------

TVL was totalSupply x price, counting OS the protocol itself holds rather than the backing that belongs to holders. It now reads Origin's protocolDailyStatDetails.tvl, converted out of the row's ETH denomination through rateETH so a Sonic product is not valued through the ETH price.

APY was a day-over-day wOS exchange-rate delta: two block-height lookups, and a single noisy day annualised. It is now measured on-chain from rebasingCreditsPerTokenHighres over 7d, via a helper shared with the other origin-* adaptors (added here as origin-ether/otokenApy.js -- identical to the copy in the origin-ether and origin-dollar branches, so whichever lands first, the rest apply cleanly).

The url was https://www.originprotocol.com/os, which returns 200 but renders a "page not found" -- OS has no product page of its own any more. Pointed at the site root, and OS gets no per-pool url.

An empty stats result previously threw a TypeError on row.tvl; it now returns no pool for that run. Chain name capitalised to Sonic.

### DIFF
--- a/src/adaptors/origin-ether/otokenApy.js
+++ b/src/adaptors/origin-ether/otokenApy.js
@@ -1,0 +1,52 @@
+/*
+ * Base APY for an Origin OToken (OETH, superOETHb, OUSD, OS), read on-chain.
+ *
+ * An OToken's rebasing multiplier is 1 / rebasingCreditsPerToken, so the ratio of that value
+ * across a window is exactly what the token rebased over it -- the realised base yield, taken
+ * straight off the token with no vault or AMO accounting involved.
+ *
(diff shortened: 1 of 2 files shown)


0xkr3p (commented):
hi @apexearth, thanks for the PR. I can see that this product has been discontinued (https://app.originprotocol.com/#/rebasing/os). Are you planning on re-enabling this product?

apexearth:
You're right we are discontinuing it. I only included it here because it had enough TVL in it still.

No problem if you don't deem it worth including.

0xkr3p:
Closing this as users can no longer deposit

PR #2901, opened 2026-08-10
termmax: dont let one failing chain discard the other three

### LINKED ISSUE #2900: termmax: no pools since it was added — one failing chain discards all four via Promise.all
`termmax` has never returned a pool. The adaptor was added on 2026-07-08 and the protocol has $32.06M TVL updating daily (`api.llama.fi/protocol/termmax`), but `yields.llama.fi/pools` has zero rows for it and it is not in `src/utils/exclude.js`.

## Cause

`apy()` fans out across the four chains and awaits them together, with no per-chain error handling (`src/adaptors/termmax/index.js:898`):

```js
async function apy() {
  const vaults = [];
  const tasks = [];
  for (const key of Object.keys(VAULTS)) {
    const task = async () => {
      const { alias, chain, chainId } = VAULTS[key];
      const vaultsOnChain = await getVaultsOnChain(chain, chainId, alias);
      for (const vault of vaultsOnChain) vaults.push(vault);
    };
    tasks.push(task());
  }
  await Promise.all(tasks);
  return vaults;
}
```

`Promise.all` rejects on the first task that throws, so one bad chain discards every vault already collected from the healthy ones. There is no partial result — the protocol reports nothing at all.

## Measured per chain

Running the adaptor with the fan-out restricted to one chain at a time:

| chain | result |
|---|---|
| ethereum | **64 pools** (incl. a $25.3M XAUt vault) |
| arbitrum | **7 pools** |
| base | **1 pool** |
| bsc | throws — `limit exceeded` / `log query range must not exceed 25 blocks` on the public BSC pool |

So 72 pools' worth of working data is being thrown away because the fourth chain fails.

To be clear about what I can and cannot show from here: the BSC failure I hit is on public RPCs and may not be the chain that fails in your infrastructure. What the numbers do establish is that the ethereum/arbitrum/base paths produce valid pools, while the protocol publishes none — which is only possible if some chain is throwing and taking the rest with it.

## Suggested fix

Catch per chain so a partial failure degrades to partial data, and throw only when every chain fails — the same shape as the `filterPools` guard in dimension-adapters #8565, so a total outage still surfaces as an error instead of an empty result.

With that change and all four chains enabled, the adaptor returns **72 pools / 438 tests passing** locally even while BSC is still failing. Forcing every chain to fail throws:

```
termmax: every chain failed -> ethereum: INJECTED | arbitrum: INJECTED | bsc: INJECTED | base: INJECTED
```

PR to follow.

0xshubhs:
Fixes #2900.

`apy()` awaited all four chains with `Promise.all` and no per-chain handling, so the first chain to throw discarded every vault already collected from the others. The protocol has $32.06M TVL updating daily but has never published a single pool.

Per chain, running the fan-out one at a time:

| chain | result |
|---|---|
| ethereum | 64 pools (incl. a $25.3M XAUt vault) |
| arbitrum | 7 pools |
| base | 1 pool |
| bsc | throws, `limit exceeded` / `log query range must not exceed 25 blocks` on public RPCs |

Catching per chain and throwing only when every chain fails, so a partial outage degrades to partial data while a total outage still surfaces as an error rather than an empty result, same shape as the `filterPools` guard in DefiLlama/dimension-adapters#8565.

Verified with all four chains enabled and BSC still failing locally:

```
$ npm run test --adapter=termmax
Tests:       438 passed, 438 total
Nb of pools: 72
```

Same on a repeat run. Forcing every chain to reject:

```
termmax: every chain failed -> ethereum: INJECTED | arbitrum: INJECTED | bsc: INJECTED | base: INJECTED
```

Worth flagging: the BSC failure I hit is on the public RPC pool and may not be what fails in your infrastructure. The fix does not depend on which chain it is; it only stops one chain from taking the rest down.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when retrieving vault data across multiple chains.
  * Successful chain results are now returned even if another chain encounters an error.
  * An error is reported only when all configured chains fail.

### DIFF
--- a/src/adaptors/termmax/index.js
+++ b/src/adaptors/termmax/index.js
@@ -898,15 +898,23 @@ async function getVaultsOnChain(chain, chainId, alias) {
 async function apy() {
   const vaults = [];
   const tasks = [];
-  for (const key of Object.keys(VAULTS)) {
+  const keys = Object.keys(VAULTS);
+  const failed = [];
+  for (const key of keys) {
     const task = async () => {
       const { alias, chain, chainId } = VAULTS[key];
-      const vaultsOnChain = await getVaultsOnChain(chain, chainId, alias);
-      for (const vault of vaultsOnChain) vaults.push(vault);
+      try {
+        const vaultsOnChain = await getVaultsOnChain(chain, chainId, alias);
+        for (const vault of vaultsOnChain) vaults.push(vault);
+      } catch (error) {
+        failed.push(`${key}: ${error.message}`);
+      }
     };
     tasks.push(task());
   }
   await Promise.all(tasks);
+  if (failed.length === keys.length)
+    throw new Error(`termmax: every chain failed -> ${failed.join(' | ')}`);
   return vaults;
 }
 


0xkr3p (approved):
lgtm - thanks @0xshubhs

PR #2926, opened 2026-08-20
add CARRY APY calculation to vault-street and change methodology

fextr:
Related TVL adapter PR: https://github.com/DefiLlama/DefiLlama-Adapters/pull/20586

Note:
If it's possible, rename the current Vault Street primeUSD page (https://defillama.com/protocol/vault-street-primeusd) to Vault Street, and show the cumulative TVL on the main page and per token in the TVL breakdown section

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking the CARRY product alongside primeUSD.
  * Improved retrieval across multiple products, allowing successful results when individual products fail.
* **Bug Fixes**
  * Improved APY accuracy using compounded annualized calculations over exact oracle intervals.
  * Preserved valid calculation windows for irregular oracle publishing schedules.
  * Restricted fallback data to unavailable oracle records while surfacing other errors.
  * Prevented results from being returned when no valid product data is available.

### DIFF
--- a/src/adaptors/vault-street-primeusd/index.js
+++ b/src/adaptors/vault-street-primeusd/index.js
@@ -4,10 +4,8 @@ const utils = require('../utils');
 const CHAIN = 'ethereum';
 
 const USDC = '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48';
-const primeUSD = '0x7ea76108975ec0998b9bc2db04b4eca986400dd7';
-const priceOracle = '0x8cda03e2004c35e07963fb792c6b7511dabee369';
 
-// oracle price is scaled to 8 decimals (primeUSD/USDC)
+// oracle prices are scaled to 8 decimals (token/USDC)
 const PRICE_DECIMALS = 8n;
 const USDC_DECIMALS = 6n;
 const LAST_PRICE_ABI =
@@ -16,7 +14,23 @@ const LAST_PRICE_ABI =
 const DAY_IN_SECONDS = 24 * 60 * 60;
 const SECONDS_IN_YEAR = 365 * DAY_IN_SECONDS;
 
-const getLastPrice = async (block) => {
+const PRODUCTS = [
+  {
+    symbol: 'primeUSD',
+    token: '0x7ea76108975ec0998b9bc2db04b4eca986400dd7',
+    priceOracle: '0x8cda03e2004c35e07963fb792c6b7511dabee369',
+    url: 'https://app.vaultstreet.com/prime-usd',
+  },
+  {
+    symbol: 'CARRY',
+    token: '0xf05f7ab9b05d9dcf99b8e9bbae8e5e4a3201d004',
+    // updates once per ~7d (staleness period 169h)
+    priceOracle: '0xd610fabab31c6d76b50a49c337fc39d6559e0e87',
+    url: 'https://app.vaultstreet.com/carry',
+  },
+];
+
+const getLastPrice = async (priceOracle, block) => {
   const { output } = await sdk.api.abi.call({
     target: priceOracle,
     abi: LAST_PRICE_ABI,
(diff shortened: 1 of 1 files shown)


src/adaptors/vault-street-primeusd/index.js
```diff
@@ -26,56 +40,83 @@ const getLastPrice = async (block) => {
...
+const getPool = async ({ symbol, token, priceOracle, url }, usdcPriceUsd) => {
+  const [supplyRes, decimalsRes, endPrice] = await Promise.all([
+    sdk.api.abi.call({ target: token, abi: 'erc20:totalSupply', chain: CHAIN }),
+    sdk.api.abi.call({ target: token, abi: 'erc20:decimals', chain: CHAIN }),
+    getLastPrice(priceOracle),
+  ]);
+
+  // sample just before the latest update landed; the oracle then reports the
+  // previous update, so the apy window is always exactly one publish interval
+  const [blockBeforeUpdate] = await utils.getBlocksByTime(
```
0xkr3p: i think rounding can help in this scenario but we should stick to min 7d, something like the below:

  Delete lines 43–47 (elapsedWholeDaysSeconds) and add a constant near line 15:
```
  const MIN_WINDOW_DAYS = 7;
```
  Lines 51–57 — take the real elapsed instead of the rounded one:
```
  const calcApyPct = (startRate, endRate, startTimestamp, endTimestamp) => {
    const elapsed = endTimestamp - startTimestamp;
    if (!(startRate > 0) || elapsed <= 0) return null;
    return (Math.pow(endRate / startRate, SECONDS_IN_YEAR / elapsed) - 1) * 100;
  };
```
  Lines 66–71 — this is the actual bug. Anchor at least MIN_WINDOW_DAYS back rather than 600 seconds:
```
    // anchor to the last publish at or before (latest - MIN_WINDOW), so the window
    // never collapses when the publisher goes off-schedule
    const [anchorBlock] = await utils.getBlocksByTime(
      [endPrice.timestamp - MIN_WINDOW_DAYS * DAY_IN_SECONDS],
      CHAIN
    );
    const startPrice = await getLastPrice(priceOracle, anchorBlock).catch(...
```
  The .catch on 72–74 stays as-is, and 88–93 needs no change — calcApyPct already receives both timestamps.

  Effect: primeUSD → ~7.0% (anchor ≈ Aug 13, elapsed ≈ 7.0d); CARRY → 9.58% (anchor Aug 12 14:29, elapsed 7.89d) instead of the current 8.72% off the 0.77d anomaly.

  And it removes both failure modes: no <12h → elapsed 0 → pool dropped, and no drift when a publish lands off-schedule, because the anchor is always ≥7 days back.

src/adaptors/vault-street-primeusd/index.js
```diff
@@ -26,56 +40,83 @@ const getLastPrice = async (block) => {
...
-      symbol: 'primeUSD',
-      tvlUsd,
-      apyBase,
-      pricePerShare,
-      underlyingTokens: [USDC],
-      token: primeUSD,
-      url: 'https://app.vaultstreet.com/prime-usd'
-    },
-  ];
+  const pools = await Promise.all(
```
0xkr3p: we could use `Promise.allSettled` so one failure does not take both pools down:
```
 const settled = await Promise.allSettled(
      PRODUCTS.map((product) => getPool(product, usdcPriceUsd))
    );
    const failed = settled
      .map((r, i) => (r.status === 'rejected' ? `${PRODUCTS[i].symbol}: ${r.reason?.message}` : null))
      .filter(Boolean);
    if (failed.length) console.log(`vault-street: ${failed.length}/${PRODUCTS.length} products failed -> ${failed.join(' | ')}`);
    if (failed.length === PRODUCTS.length) throw new Error(`vault-street: every product failed -> ${failed.join(' | ')}`);
    return settled.filter((r) => r.status === 'fulfilled').map((r) => r.value).filter(Boolean);
```

0xkr3p (commented):
gm @fextr, thanks for the PR - I've left a couple of comments

src/adaptors/vault-street-primeusd/index.js
fextr: thx, done

src/adaptors/vault-street-primeusd/index.js
fextr: done

fextr:
> gm @fextr, thanks for the PR - I've left a couple of comments

@0xkr3p thx for the review! applied fixes

0xkr3p (approved):
thanks for the fast changes @fextr approved

PR #2927, opened 2026-08-20
hrusd: add HRUSD/USDC Uniswap V3 staking pool

teamskew:
Adds the HRUSD/USDC Uniswap V3 staking pool. HRUSD is already listed on the TVL side
(`hrusd`, protocol id 8458) and as a pegged asset (id 433).

### Where the APY comes from

The reward rate is a single on-chain parameter, not an off-chain figure:
`aprBps()` returns `1200` on both V3LPStakingRewards deployments, against
`BPS_DENOMINATOR() = 10000` — a flat 12.00% APR accruing linearly on each stake's
`principalHrusd`. The adapter reads both deployments and takes the lower of the two, per
the unboosted-value rule.

No haircut is applied for early exit. The exit fee is set to zero: `ExitFeesPaid` has never
fired on the current deployment, and over the legacy deployment's entire lifetime it fired
24 times for 0.977958 HRUSD in total. Unstaking is gated by a 24h delay
(`unstakePeriod() = 86400`), not by a penalty on accrued rewards.

### How TVL is computed

Rewards accrue on `principalHrusd`, the HRUSD-denominated value of a staked position, so
that is what the pool's TVL measures. The adapter enumerates the position NFTs escrowed by
the two staking contracts through the Uniswap position manager, calls `previewAmounts` on
each, and sums the principals — currently 27 positions totalling 41,172.68 HRUSD.

HRUSD has no price feed on DefiLlama (`coins.llama.fi` returns an empty object for it), so
it is valued against USDC from the HRUSD/USDC pool itself, using a time-averaged tick from
`observe()` over 1800s — the same window the staking contracts use for their own principal
valuation (`twapWindowSeconds() == 1800`). A spot `slot0()` read would let a swap at the
sampled block move the reported TVL; `observe` reverts if the pool's observation cardinality
does not cover the window, which fails the run rather than publishing a manipulable number.

As a cross-check, the USDC leg of those same positions sums to 9,589.10, which matches the
`pool2` figure produced by the existing `hrusd` entry in DefiLlama-Adapters.

### Note on the TVL difference vs the protocol page

This pool reports ~$41.2k while `defillama.com/protocol/hrusd` shows ~$10.2k. That is
expected and not a discrepancy: the protocol-side config deliberately counts only the USDC
side, because the HRUSD it is paired against is backed by that same USDC and counting both
would double count. A yield pool's TVL is the full value of the staked position, so both
legs are included here.

### Note on the reward token

Rewards are paid in HRUSD, funded into the staking contracts up front via `fund()`. Flagging
this explicitly against the "omit any pre-mined rewards" rule: the rewards are pre-funded
inventory rather than emissions minted on claim. The same balance is excluded from HRUSD's
circulating supply in peggedassets-server (PR #901), on the grounds that it has not been
distributed to any holder — `totalOwed()` currently reads 0. If you would rather this pool
carry no `apyReward` on that basis, say so and I will drop it to a TVL-only entry.

HRUSD's only venue is the Uniswap V3 pool above; it is not listed on CoinGecko or CMC.

### Validation

```
$ npm run test --adapter=hrusd

Tests: 14 passed, 14 total
```

Output:

```json
{
  "pool": "0xae2e818a18e95212fae482e2180bc89546393dc9-base",
  "chain": "Base",
  "project": "hrusd",
  "symbol": "HRUSD-USDC",
  "tvlUsd": 41146.09,
  "apyReward": 12,
  "rewardTokens": ["0x5587AD03F9565F1B86cfA51a3C744Bcc4039dAf0"],
  "underlyingTokens": ["0x5587AD03F9565F1B86cfA51a3C744Bcc4039dAf0", "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913"],
  "token": "0xAE2E818A18e95212FAE482e2180bC89546393DC9",
  "poolMeta": "Uniswap V3 LP staking"
}
```

One new file, no dependency or lockfile changes.

Both of CodeRabbit's findings are addressed in the second commit: the TWAP change above, and
dropping `permitFailure` from the `previewAmounts` multiCall so an incomplete set of position
previews fails the run instead of silently reporting partial TVL.

Unrelated, but worth flagging while I was here: `beforeTests.js` resolves the adapter path
with `cwd.includes('src/adaptors')`, which never matches on Windows because the path uses
backslashes, so the runner looks under `src/adaptors/src/adaptors/`. Happy to open a separate
PR for that if useful — this one leaves it untouched.

## Summary by CodeRabbit

* **New Features**
  * Added tracking for pending HRUSD staking rewards.
  * Added base APY calculations using recent trading fees and current pool value.
  * Added separate HRUSD and USDC pricing.
  * Improved position reporting to include staked principal and accrued rewards.
  * Added support for HRUSD staking positions across multiple deployments.
  * Added HRUSD pricing, total value locked, reward, and underlying token details.
  * Added unboosted APY calculations based on staking rewards.
  * Added protocol metadata and configuration for HRUSD staking.
* **Bug Fixes**
  * Improved valuation reliability by rejecting incomplete pricing data and unavailable liquidity windows.

### DIFF
--- a/src/adaptors/hrusd/index.js
+++ b/src/adaptors/hrusd/index.js
@@ -25,7 +25,15 @@ const ABI = {
     'function previewAmounts(uint256 tokenId) view returns (uint256 amt0, uint256 amt1, uint256 principalHrusd)',
   observe:
     'function observe(uint32[] secondsAgos) view returns (int56[] tickCumulatives, uint160[] secondsPerLiquidityCumulativeX128)',
+  pendingRewards: 'function pendingRewards(uint256 tokenId) view returns (uint256)',
+  feeGrowth0: 'uint256:feeGrowthGlobal0X128',
+  feeGrowth1: 'uint256:feeGrowthGlobal1X128',
+  liquidity: 'uint128:liquidity',
 };
+
+// Base produces a block roughly every 2s.
+const BLOCKS_PER_DAY = 43200;
+const Q128 = 2n ** 128n;
 
 // Same window the staking contracts use for their own principal valuation
 // (`twapWindowSeconds() == 1800`).
@@ -37,7 +45,7 @@ const TWAP_WINDOW = 1800;
 // block cannot move the reported TVL. `observe` reverts if the pool's observation
 // cardinality does not cover the window, which fails the run rather than reporting a
 // manipulable number.
-const getHrusdPrice = async (api) => {
+const getPrices = async (api) => {
   const [observation, usdc] = await Promise.all([
     api.call({ abi: ABI.observe, target: POOL, params: [[TWAP_WINDOW, 0]] }),
     utils.getData(`https://coins.llama.fi/prices/current/${CHAIN}:${USDC}`),
@@ -51,12 +59,14 @@ const getHrusdPrice = async (api) => {
   if (!Number.isFinite(avgTick)) throw new Error('hrusd: invalid TWAP observation');
 
   // HRUSD is token0, USDC is token1, both 6 decimals, so the tick needs no decimal shift
-  return 1.0001 ** avgTick * usdcPrice;
+  return { hrusdPrice: 1.0001 ** avgTick * usdcPrice, usdcPrice };
 };
 
 // Rewards accrue linearly on `principalHrusd`, the HRUSD-denominated value of each
-// staked position, so the pool's TVL is the sum of those principals.
-const getStakedPrincipal = async (api) => {
+// staked position, so the pool's TVL is the sum of those principals. The same walk over
+// the staked positions yields `owed`, the rewards already accrued to stakers, which the
+// reward runway below has to net out of the contracts' balance.
+const getStakedPositions = async (api) => {
   const counts = await api.multiCall({
     abi: ABI.balanceOf,
     target: POSITION_MANAGER,
@@ -68,7 +78,7 @@ const getStakedPrincipal = async (api) => {
     for (let j = 0; j < Number(counts[i]); j += 1)
       idCalls.push({ owner, params: [owner, j] });
   });
-  if (!idCalls.length) return 0;
+  if (!idCalls.length) return { principal: 0, owed: 0 };
 
   const tokenIds = await api.multiCall({
     abi: ABI.tokenOfOwnerByIndex,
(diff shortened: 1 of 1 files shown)


0xkr3p (commented):
hi @teamskew, thanks for the PR - could we implement a sanity check around the rewards, e.g. if aprBps stays at 1200, the adapter continues to post 12% a sense check with totalOwed / balance could be useful to prevent this from happening. Also is this missing `apyBase` if i's a uniswap v3 pool?

teamskew:
Thanks @0xkr3p — both points were fair, and both are addressed in `77b3124`.

### Sanity check on the rewards

You were right that `aprBps` on its own is not evidence of anything: it is an owner-set
parameter, so the adapter would have kept posting 12% after the reward pool ran dry.

The adapter now does exactly the `totalOwed` / balance check you suggested. It reads the
HRUSD balance held by both V3LPStakingRewards deployments, then walks every staked position
and sums `pendingRewards(tokenId)` — the rewards already accrued to stakers. Only the
remainder can fund future accrual, and if nothing is left unspoken for, `apyReward` is
reported as 0 rather than the nominal rate.

At the current block:

| | HRUSD |
|---|---|
| Held by the two staking contracts | 509.70 |
| Already accrued to stakers (27 positions) | 192.10 |
| Unspoken for | 317.60 |
| Accruing per day (12% on 41,172.68 principal) | 13.54 |

which is roughly 23 days of runway.

For context on how that gets replenished: the protocol targets a ~30 day buffer, topping the
contracts up from revenue earned off-chain and converted into HRUSD by a bot. That process is
not verifiable from here, so the adapter does not rely on it — it measures what the contracts
actually hold on every run and gates the reward APY on that. If the top-ups stop, the number
goes to zero on its own.

Happy to make the gate stricter — e.g. require N days of runway rather than merely non-zero —
if you would prefer a floor.

### apyBase

Added, computed rather than assumed: the delta in the pool's `feeGrowthGlobal0X128` /
`feeGrowthGlobal1X128` over ~24h of blocks, multiplied by in-range `liquidity`, valued at the
pool price and annualised against the pool's own TVL.

It currently comes out at **0.00%**, and that is a real measurement rather than a placeholder:
both fee accumulators are byte-identical to their values ~43,200 blocks ago. The pool has had
essentially no trading — GeckoTerminal puts the last 24h at 1 swap for $2.13 against ~$50k of
liquidity. So the fee component is genuinely nil today; the code will pick it up if that
changes.

### Validation

```
$ npm run test --adapter=hrusd
Tests: 14 passed, 14 total
```

```json
{
  "tvlUsd": 41159.03,
  "apyBase": 0,
  "apyReward": 12,
  ...
}
```

teamskew:
@0xkr3p some context on why the rate holds, on top of the on-chain gate already pushed in
`77b3124`.

The rewards are not funded from a pre-mined allocation that depletes. HyperRoute runs a
delta-neutral strategy on a centralised venue — similar in shape to Ethena's — whose revenue
accrues in USDC. A bot converts that USDC into HRUSD and forwards it to the staking
contracts, so the reward inventory is topped up out of running revenue rather than drawn
down from a fixed pot. That is why `aprBps` has stayed at 1200 rather than being ratcheted
down.

The part of that which is actually checkable is the inventory itself, so here it is.
Reconstructing every HRUSD `Transfer` into and out of both V3LPStakingRewards deployments
since they were created:

| | legacy | current |
|---|---|---|
| Transfers | 142 | 21 |
| First block | 45,251,414 | 49,605,254 |
| Lowest balance after initial funding | 89.64 | 316.66 |
| Times the balance reached zero | 0 | 0 |

The inventory has never been exhausted. Both contracts open with a 1.00 HRUSD test transfer,
which is the only reason a sub-1 figure appears anywhere in their history.

To be explicit about the limits of that: the top-up process itself is off-chain and you have
no way to verify it from this repo, and neither do I from an adapter. That is exactly why the
adapter does not assume it. It nets `pendingRewards` out of the contracts' HRUSD balance on
every run and reports `apyReward: 0` if nothing is left unspoken for — so if the strategy
stops funding, the number falls away on its own rather than sitting at 12% indefinitely.

Current state: 509.70 held, 192.10 already accrued to stakers, 317.60 unspoken for, against
13.54 HRUSD/day of accrual.

teamskew:
Both remaining CodeRabbit findings are addressed in `cb3a271`.

### Liquidity behind the TWAP

The observation now yields more than a tick. `secondsPerLiquidityCumulativeX128` is read
alongside `tickCumulatives` and turned into the harmonic mean of in-range liquidity over the
window; if that is degenerate the tick describes an empty pool and the run fails instead of
pricing TVL off it.

I did not add a floor above zero, deliberately. Any number I picked would be arbitrary, and
HRUSD has no second venue to fall back on — so a threshold set too high would delist the pool
rather than protect it, and one set low enough to be safe would not stop a determined
manipulator anyway. The guard is therefore limited to the case it can justify: no liquidity at
all behind the window. Happy to add a concrete floor if you have a house value for it.

### Fees and changing liquidity

That one was simply a bug. `feeGrowthGlobal` rises by fee/liquidity at each swap, so
multiplying its delta by a single liquidity reading is only right while liquidity holds still
across the window.

Rather than reconstructing per-swap fee growth, the adapter now reads the pool's `Swap`
events over the window and takes the fee tier out of each swap's input amount. That is exact
by construction and never touches liquidity, so the failure mode disappears instead of being
approximated around.

It also turns out to be more sensitive than what it replaced. `apyBase` now reports
0.00094% where the accumulator method reported a flat 0: the pool's one swap in the window,
$2.13 against ~$50.7k of liquidity at the 0.05% tier, works out to roughly $0.001/day, which
annualises to about that. Still a rounding error as a yield, but it is now a measured one.

### Validation

```
$ npm run test --adapter=hrusd
Tests: 14 passed, 14 total
```

```json
{ "tvlUsd": 41160.22, "apyBase": 0.00094, "apyReward": 12 }
```

0xkr3p (approved):
approved, thanks for the fast changes @teamskew

PR #2937, opened 2026-08-23
feat: add Ammalgam DLEX yield pools

duelinggalois:
## Summary
- Add a timetravel-capable Ammalgam DLEX Yields adapter for Ethereum USDC-WETH and USDC-USDT pools.
- Keep the adapter self-contained in `src/adaptors/ammalgam-dlex/index.js` and reuse the existing DefiLlama SDK/utils stack.
- Match Uniswap v2-style fee-yield rows by reporting `apyBase`, `apyBase7d`, `volumeUsd1d`, `volumeUsd7d`, `underlyingTokens`, `token`, and pool URLs.
- Compute reserve-only pool TVL from `getReserves()`.
- Compute base APY from dynamic swap fees plus LP borrow interest from `InterestAccrued` reserve growth.

## Context
DefiLlama closed the separate Ammalgam DEX TVL PR because the existing DEX and DLEX pages will be combined for the same factory/pair. This PR reports the pools under Ammalgam DLEX instead of introducing an `ammalgam-dex` yield project.

## Test Plan
- `npm run test --adapter=ammalgam-dlex --timestamp=1787950000`
- `npx prettier --check src/adaptors/ammalgam-dlex/index.js`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking Ammalgam DLEX yield markets on Ethereum.
  * Added APY reporting for USDC/WETH and USDC/USDT pools, including daily and weekly rates.
  * Added historical data support with USD-denominated calculations and timetravel capabilities.

### DIFF
--- a/src/adaptors/ammalgam-dlex/index.js
+++ b/src/adaptors/ammalgam-dlex/index.js
@@ -1,0 +1,408 @@
+const sdk = require('@defillama/sdk');
+const {
+  utils: { formatUnits },
+} = require('ethers');
+const utils = require('../utils');
+
+const PROJECT = 'ammalgam-dlex';
+const CHAIN = 'ethereum';
+const DISPLAY_CHAIN = utils.formatChain(CHAIN);
+const SECONDS_PER_DAY = 24 * 60 * 60;
+const DAYS_PER_YEAR = 365;
+
+const token = (symbol, address, decimals) => ({
+  symbol,
+  address: address.toLowerCase(),
+  decimals,
+});
+
+const USDC = token('USDC', '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', 6);
+const USDT = token('USDT', '0xdac17f958d2ee523a2206206994597c13d831ec7', 6);
+const WETH = token('WETH', '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', 18);
+
+const TOKENS = [USDC, USDT, WETH];
+const POOLS = [
+  {
+    pair: '0x728fd0a966b993fe518b00122d51e494f99abd6a',
+    fromBlock: 25481998,
+    fromTimestamp: 1783443275,
+    token: '0x72d015e116a965ed022c57dbc18a91f0bd329b65',
+    tokens: [USDC, WETH],
+  },
+  {
+    pair: '0xf53d16bc876212ae501cccc1949d73bb55be4b0e',
+    fromBlock: 25843244,
+    fromTimestamp: 1787793803,
+    token: '0xb2e1c416b97613a30a69c5b12ec369d8e986d5ac',
+    tokens: [USDC, USDT],
+  },
+].map((pool) => ({
+  ...pool,
+  pair: pool.pair.toLowerCase(),
+  token: pool.token.toLowerCase(),
+}));
+
+const PRICE_KEYS = TOKENS.map((token) => `${CHAIN}:${token.address}`);
+
+const GET_RESERVES_ABI =
+  'function getReserves() view returns (uint112 reserveXAssets, uint112 reserveYAssets, uint32 lastTimestamp)';
+const SWAP_EVENT =
+  'event Swap(address indexed sender, uint256 amountXIn, uint256 amountYIn, uint256 amountXOut, uint256 amountYOut, address indexed to)';
+const SYNC_EVENT = 'event Sync(uint256 reserveXAssets, uint256 reserveYAssets)';
+const INTEREST_ACCRUED_EVENT =
+  'event InterestAccrued(uint256 reserveXAssets, uint256 reserveYAssets, uint112 depositXAssets, uint112 depositYAssets, uint112 borrowLAssets, uint112 borrowXAssets, uint112 borrowYAssets)';
+
+const toBigInt = (amount) => BigInt(amount.toString());
+
+const ceilDiv = (numerator, denominator) =>
+  (numerator + denominator - 1n) / denominator;
+
+const amountInForNoFeeSwap = (reserveIn, reserveOut, amountOut) => {
+  if (amountOut === 0n) return 0n;
+  if (reserveIn === 0n || reserveOut <= amountOut) return null;
+
+  return ceilDiv(reserveIn * amountOut, reserveOut - amountOut);
+};
+
+const getReserveState = (reserves) => ({
+  reserveXAssets: toBigInt(reserves.reserveXAssets ?? reserves[0]),
+  reserveYAssets: toBigInt(reserves.reserveYAssets ?? reserves[1]),
+});
+
+const emptyReserveState = () => ({
+  reserveXAssets: 0n,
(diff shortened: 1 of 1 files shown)


duelinggalois:
Hey @bheluga anything I can do on my end to make this easy to review?

src/adaptors/ammalgam-dlex/index.js
```diff
@@ -0,0 +1,349 @@
...
+const buildPool = async (pool, blocks, prices) => {
+  if (blocks.endBlock < pool.fromBlock) return null;
+
+  const [reserves, dailyYield, weeklyYield] = await Promise.all([
+    getReservesAtBlock(pool, blocks.endBlock),
+    calculateWindowYield(pool, blocks.dayStartBlock, blocks.endBlock),
+    calculateWindowYield(pool, blocks.weekStartBlock, blocks.endBlock),
+  ]);
+  const tvlUsd = toUsd(
+    { x: reserves.reserveXAssets, y: reserves.reserveYAssets },
```
0xkr3p, line 338: i think the tvl is overstated by counting borrowed capital twice, we can use balance of for both underlying assets to get tvl at end block

src/adaptors/ammalgam-dlex/index.js
```diff
@@ -0,0 +1,349 @@
...
+    prices
+  );
+
+  return {
+    pool: `${pool.pair}-${CHAIN}`,
+    chain: DISPLAY_CHAIN,
+    project: PROJECT,
+    symbol: pool.tokens.map((token) => token.symbol).join('-'),
+    tvlUsd,
+    apyBase: tvlUsd > 0 ? (dailyYieldUsd / tvlUsd) * DAYS_PER_YEAR * 100 : 0,
```
0xkr3p: fallback to null rather than 0, same for 7d below

src/adaptors/ammalgam-dlex/index.js
```diff
@@ -0,0 +1,349 @@
...
+  };
+};
+
+const apy = async (timestampArg = null) => {
+  const timestamp = normalizeTimestamp(timestampArg);
+  const [blocks, prices] = await Promise.all([
+    getBlocks(timestamp),
+    getHistoricalPrices(timestamp),
+  ]);
+  const pools = await Promise.all(
```
0xkr3p: we could use Promise.allSettled to help keep adapter running if 1 pool drops

0xkr3p (commented):
gm @duelinggalois, thanks for the pr - i've left a few comments to resolve before we can merge

src/adaptors/ammalgam-dlex/index.js
duelinggalois: I am not sure I understand why borrowed capital would be counted twice. The reserves can be borrowed, so they can be higher then the actual token balances. An example being [USDC-USDT](https://etherscan.io/address/0xf53D16Bc876212Ae501cCCc1949d73BB55Be4b0E) pair we created this week has $16k in token holdings for both tokens, and the reserves are roughly $36k for both tokens.
```
{
  _reserveXAssets (uint112) : 18026845142
  _reserveYAssets (uint112) : 18026844499
  _lastUpdateTimestamp (uint32) : 1787984519
}
```

My goal was to compute the apy correctly, lowering the tvl to reflect the lent out assets then inflates the yield for `apyBase` and `apyBase7d`. This also creates a potential confusion when comparing to a [uni v2 pool](https://defillama.com/yields/pool/702fff16-4bfa-4c1f-8c02-e0ea0f5ecde6?supplyMagnitude=true&supplyAvg7d=true) if a different definition of reserves is used to reflect the borrowed assets. 

I am happy to make changes you suggest. Perhaps a second variable could be used for final returned `tvlUsd` but borrowed X and Y will be sourced from both the reserves and deposited X and Y, so there are other questions of how exactly to measure how much of borrowed X or Y comes from the reserves vs the deposits. 

What would you suggest?

0xkr3p: apologies for the confusion re: counted capital twice, borrows should not be included in the tvl number as this overstates, tvl is reserves + deposits - borrows. yes I think something like this would work:
```
   const [reserves, balances, dailyYield, weeklyYield] = await Promise.all([
     getReservesAtBlock(pool, blocks.endBlock),   // APY denominator — unchanged
     getBalancesAtBlock(pool, blocks.endBlock),   // new: erc20:balanceOf per underlying
     calculateWindowYield(pool, blocks.dayStartBlock, blocks.endBlock),
     calculateWindowYield(pool, blocks.weekStartBlock, blocks.endBlock),
   ]);
  
   const reservesUsd = toUsd({ x: reserves.reserveXAssets, y: reserves.reserveYAssets }, pool, prices);
   const tvlUsd      = toUsd(balances, pool, prices);
  
   apyBase:   (dailyYieldUsd  / reservesUsd) * (365 / dailyYield.elapsedDays)  * 100,
   apyBase7d: (weeklyYieldUsd / reservesUsd) * (365 / weeklyYield.elapsedDays) * 100,
```

src/adaptors/ammalgam-dlex/index.js
duelinggalois: Fixed in 657d15f9: per-pool builds now use `Promise.allSettled`, then keep fulfilled/finite pools so one pool failure does not reject the whole adapter.

src/adaptors/ammalgam-dlex/index.js
duelinggalois: Fixed in 657d15f9: `apyBase` and `apyBase7d` now fall back to `null` when TVL is not positive.

src/adaptors/ammalgam-dlex/index.js
duelinggalois: Implemented in 3a9093c2. `tvlUsd` now uses each underlying token balance held by the pair, while `apyBase` and `apyBase7d` keep using reserves as the denominator. The APY windows also annualize by elapsed days so new pools do not treat a partial lookback as a full 1d/7d window.

src/adaptors/ammalgam-dlex/index.js
duelinggalois: Current head is 3a9093c2 after the amend. This remains fixed there: `apyBase` and `apyBase7d` return `null` when the denominator/window is not valid instead of falling back to `0`.

src/adaptors/ammalgam-dlex/index.js
duelinggalois: Current head is 3a9093c2 after the amend. This remains fixed there: pool builds use `Promise.allSettled` and only fulfilled finite pools are returned, so one pool failure does not reject the full adapter.

duelinggalois:
> gm @duelinggalois, thanks for the pr - i've left a few comments to resolve before we can merge

@0xkr3p I have made changes for two of your comments and left a reply  and question to your tvl comment.

0xkr3p (approved):
approved thanks for the changes @duelinggalois

PR #2944, opened 2026-08-26
Add FWB (Finance Without Borders) yield adapter

fwb030101:
Adds a yield adapter for FWB (Finance Without Borders), protocolId 8069,
already listed on the TVL side as `fwb`.

1. What it is. A fixed-rate, fixed-term vault: users deposit USDT, WETH or a
   wrapped BTC asset for 30, 90, 180 or 360 days. The rate is locked at deposit
   time and does not float afterwards.

2. Pool identity. The protocol issues no ERC-20 receipt token — a position
   exists only as an entry in the vault's internal accounting. There is
   therefore no receipt address to build the pool id from, so we use
   `${vaultAddress}-${assetAddress}-${chain}`.

3. APY. A pool has no single rate: the rate depends on the term chosen by the
   depositor (currently 6/8/10/12% for USDT across 30/90/180/360 days). We
   report the 360-day rate and state this explicitly in `poolMeta`. This mirrors
   how fixed-maturity yield is displayed elsewhere on DefiLlama, where the rate
   held to maturity is shown even though exiting early through the market can
   return less.

4. Data sources. On-chain reads only, no APIs:
   - `getSupportedTokens()` — accepted assets
   - `activePrincipalByToken(token)` — active principal
   - `getTokenUsdValue18(token, amount)` — USD value, 18 decimals
   - `currentAprBps(token, term)` — rate in basis points

5. Deployments (same contract, five chains):
   - Ethereum  0x1Ef96B8fad9aE983E60610C4ba13536606B5c477
   - BNB Chain 0x18A021d1c89Af87AaeD266B2C58dD16855Ad3702
   - Polygon   0xd17127796D46c1588550Df783FCfE3D08ef8F6c0
   - Arbitrum  0xF5d84413f2cd33d6d473BA9D0c665a73472d8fC7
   - Base      0x199180dfbACEE5c204Db4E803A92a9D3A9Db4d1F

6. Output. 15 pools, aggregate TVL ~$52k, which matches the TVL currently
   reported for the protocol.

Site: https://finwb.xyz

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking FWB fixed-term DeFi yield pools.
  * Added pool data across Ethereum, BSC, Polygon, Arbitrum, and Base.
  * Included normalized token details, TVL, APY, underlying assets, metadata, and protocol links.
  * Added clearer pool labeling for 360-day fixed-term products and improved links to pool details.

### DIFF
--- a/src/adaptors/fwb/index.js
+++ b/src/adaptors/fwb/index.js
@@ -1,19 +1,18 @@
-// Адаптер доходности FWB для DeFiLlama (репозиторий DefiLlama/yield-server).
-// Путь в репозитории: src/adaptors/fwb/index.js
+// FWB yield adapter for DefiLlama (DefiLlama/yield-server).
 //
-// Протокол: фиксированная ставка на фиксированный срок (30, 90, 180, 360 дней).
-// Расписок ERC-20 за депозит нет — позиция живёт во внутреннем учёте контракта,
-// поэтому идентификатор пула строим из адреса хранилища и адреса актива.
+// Protocol: fixed rate for a fixed term (30, 90, 180 or 360 days).
+// There is no ERC-20 receipt token — a deposit lives in the vault's internal
+// accounting, so the pool id is built from the vault address and the asset address.
 //
-// Что откуда берём (всё — вызовы чтения в контракте, без внешних API):
-//   getSupportedTokens()                      список принимаемых активов
-//   activePrincipalByToken(token)             тело активных депозитов
-//   getTokenUsdValue18(token, amount)         оценка в долларах, 18 знаков
-//   currentAprBps(token, term)                ставка в базисных пунктах
+// Where the data comes from (all on-chain reads, no external APIs):
+//   getSupportedTokens()                accepted assets
+//   activePrincipalByToken(token)       principal of active deposits
+//   getTokenUsdValue18(token, amount)   USD value, 18 decimals
+//   currentAprBps(token, term)          rate in basis points
 //
-// APY подаём для срока 360 дней и указываем это в poolMeta: единой ставки
-// у пула нет, срок выбирает вкладчик. Тот же подход у рынков с фиксированной
-// доходностью до погашения.
+// APY is reported for the 360-day term and poolMeta says so: there is no single
+// rate for a pool, the term is chosen by the depositor. Same approach as other
+// fixed-maturity markets.
 
 const sdk = require('@defillama/sdk');
 const utils = require('../utils');
@@ -21,7 +20,7 @@ const utils = require('../utils');
 const PROJECT = 'fwb';
 const SITE = 'https://finwb.xyz';
 
-// Срок, ставку за который показываем.
+// The term whose rate we report.
 const TERM_DAYS = 360;
 
 const VAULTS = {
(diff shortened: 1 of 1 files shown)


src/adaptors/fwb/index.js
```diff
@@ -0,0 +1,126 @@
...
+
+  return tokens.map((token, i) => {
+    const tvlUsd = Number(values.output[i].output) / WAD;
+    const apyBase = Number(aprs.output[i].output) / BPS;
+
+    return {
+      pool: `${target}-${token}-${chain}`.toLowerCase(),
+      chain: utils.formatChain(chain),
+      project: PROJECT,
+      symbol: utils.formatSymbol(normaliseSymbol(symbols.output[i].output)),
```
0xkr3p: remove format symbol

src/adaptors/fwb/index.js
```diff
@@ -0,0 +1,126 @@
...
+      tvlUsd,
+      apyBase,
+      underlyingTokens: [token],
+      // Единой ставки у пула нет — она зависит от выбранного вкладчиком срока.
+      // Показываем ставку за 360 дней и говорим об этом прямо.
+      poolMeta: `Fixed term 30/90/180/${TERM_DAYS}d, APY shown for ${TERM_DAYS}d`,
```
0xkr3p: reduce pool meta, the text is too long

src/adaptors/fwb/index.js
```diff
@@ -0,0 +1,126 @@
...
+      url: SITE,
```
0xkr3p: add per pool url

src/adaptors/fwb/index.js
0xkr3p: add `token` property with `null` value if there isn't a receipt token

0xkr3p: probably `Fixed term 360d` if this is the pool that is represented

0xkr3p (commented):
hi @fwb030101, thanks for the pr - i've left a few comments. additionally, can we implement pricing via defillama api and replace or remove all non-english comments

fwb030101:
Thanks for the review — all five points are addressed in b324b8d.

- **format symbol** — removed. Kept only the U+20AE normalisation: the on-chain symbol is `USD₮0` with `₮` (U+20AE), and `formatSymbol` stripped that character, turning it into `USD0`. It now comes out as `USDT0`.
- **poolMeta** — shortened to `Fixed term 360d`, as you suggested.
- **token** — added as `null`. There is no receipt token: a deposit is tracked in the vault's internal accounting.
- **per pool url** — this one I could not do properly, and I'd rather say so than fake it. The app has no per-pool page: chain, asset and term are all selected in the UI after the wallet is connected, and the frontend ignores query parameters such as `?chain=ethereum` (I checked). The url now points at the deposit page `/cabinet` rather than the site root — the closest thing available. If you'd prefer the root, or want the field dropped, just say so.

I also translated all code comments to English — they were in Russian, my oversight.

Tests pass: 96/96, 15 pools.

0xkr3p:
hi @fwb030101 thanks for those changes,  I'm keen to understand more about the protocol, I can't see any of the TVL in the vaults addresses on chain?

fwb030101:
Correct — the vault addresses are not custody addresses, so their on-chain token
balances are ~0 by design.

1. The vault keeps per-token accounting of active user principal on-chain;
   deposits themselves are moved out for managed fixed-yield operations.
2. This is the methodology already accepted for the FWB TVL adapter
   (protocolId 8069, module finwb/index.js) and published on the protocol page:
   "FWB TVL is calculated from active user principal tracked on-chain by each
   deployed FWB contract via getProtocolTvlTokensAndAmounts(). User deposits are
   transferred to protocol payout wallets for managed fixed-yield operations,
   while the contracts keep active principal accounting per supported token."
3. This yield adapter reads the same on-chain source — activePrincipalByToken(token),
   valued with getTokenUsdValue18(). The sum across the five chains matches the TVL
   shown on the protocol page: $52,601 on 2026-09-07.
4. So tvlUsd here is outstanding user principal under the already-accepted
   methodology, not the ERC-20 balance of the vault address.

Happy to clarify anything else on the adapter side.

0xkr3p (approved):
approved - thanks for the info

PR #2949, opened 2026-08-27
d2-finance: on-chain epoch-based APR 

HWxFrank:
D2 vaults trade in discrete epochs (FundsCustodied → FundsReturned); NAV (pricePerShare, ERC-4626) only steps
at settlement, so the previous 30d-lookback read 0% for most pools whenever no epoch settled inside the
window.

Changes:
- apyBase = cumulative NAV growth annualized over effective trading days (sum of custody→return spans from on-chain epoch events) — matches realized, net-of-fees strategy performance - apyBaseInception = same NAV growth annualized over calendar days since deploy (deposit-and-hold basis); also serves as fallback if the event scan fails
- Epoch events fetched with one fan-in eth_getLogs per chain (all vaults + both topics, adaptive range
chunking); pps/TVL unchanged, fully on-chain
-  Pool list trimmed to the 7 actively marketed strategies (legacy/wound-down vaults removed; protocol TVL is unaffected — tracked by the TVL adapter)

Tested: npm run test --adapter=d2-finance — 62/62 passing; APRs verified against on-chain NAV and d2.finance

## Summary by CodeRabbit

- **New Features**
  - Updated available D2 Finance strategies to focus on actively marketed Base, Arbitrum, and Hyperliquid vaults.
  - Added more accurate performance metrics based on completed trading days and annualized growth since deployment.
  - Added inception-based performance reporting for additional context.

- **Bug Fixes**
  - Improved historical performance calculations for vaults with limited or incomplete epoch data.
  - Enhanced data retrieval reliability through retries and fallback handling.

### DIFF
--- a/src/adaptors/d2-finance/index.js
+++ b/src/adaptors/d2-finance/index.js
@@ -1,60 +1,46 @@
+const axios = require('axios');
 const sdk = require('@defillama/sdk');
 
 const utils = require('../utils');
 
+// Only actively marketed strategies are listed on the yields page; protocol
+// TVL is tracked separately (DefiLlama-Adapters) and counts every vault.
 const config = {
-  ethereum: ['0x07Dff4087b43c4A759f4Fc69511c26d51929dAF4'],
   base: [
-    '0x6c05A7d2c24B48fC3C615D294fEc2eB068548897',
-    '0x2406aacbdF8463176DeB285AdAa81768415B6c7E',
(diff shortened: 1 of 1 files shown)


0xkr3p (commented):
hi @HWxFrank, thanks for the pr - `apyBaseFromInception` is going to be deprecated, using `calendarApr` on apyBase does not show an accurate representation of current yield on offer and can be slow to react as the duration progresses. I'll catch up with the team on how best to measure this and get back to you.

PR #2951, opened 2026-08-28
drop pools holding implausibly more tvl than their own protocol

### LINKED ISSUE #2950: a pool can publish more tvl than its whole protocol: aerodrome-slipstream WETH-SAND reads $31.4b on a pool holding 2.15 weth, and boundaries.tvlUsdUI.ub is dead code
`/pools` currently returns this:

```
project  aerodrome-slipstream
chain    Base
symbol   WETH-SAND
pool     6a0fa42d-494e-44d2-adfb-39b3a8eacb5f
tvlUsd   31,431,600,393
```

that is 17.11% of the $183,753,699,148 total across all 17,274 pools, sitting on a pool that holds
$5,352.

the pool is `0xa69cdf524d072c366fd050957880552916d29405` (slipstream CL100 WETH/SAND, resolved via
`getPool` on the CL factory). balances on base right now:

```
WETH  2.147927                   -> $5,352 at the coins.llama.fi price for base:0x4200...0006
SAND  26,219,140,097,322.51
```

so the published number implies SAND at $0.0011988/token. `totalSupply` is 327,574,531,179,830, which
puts the implied market cap at $392.7b. `aerodrome-slipstream` protocol tvl on api.llama.fi is
$129,865,242, so the single pool is 242x its own protocol, and 117x the whole aerodrome parent
(v1 $121.2m + slipstream $129.9m + ignition $17.3m = $268.3m).

coins.llama.fi has no price for that token at all, so there is nothing upstream to correct:

```
/prices/current/base:0xac531eb26ca1d21b85126de8fb87e80e09002dcf   -> {"coins":{}}
/prices/first/base:0xac531eb2...                                  -> {"coins":{}}
/prices/historical/<2026-08-20..24>/base:0xac531eb2...            -> {"coins":{}}
```

`/chart/6a0fa42d-494e-44d2-adfb-39b3a8eacb5f`:

```
2026-08-20            114,757
2026-08-21            121,012
2026-08-22     31,431,600,393
```

one 260,000x step, and 08-22 is the last point on the chart. sibling pools from the same adapter
(WETH-USDC, WETH-SUPEROETHB) have points through 2026-08-28, so this pool stopped updating right
after the bad value landed and `/pools` has been serving it for six days since. `boundaries.age` is
7 so it does age out on its own, but the same sequence repeats for any pool that glitches once.

## why nothing caught it

`src/utils/exclude.js`:

```js
// we only get pools for the UI with a tvlUsd of minimum $10k and max ($20 billion)
tvlUsdUI: { lb: 10000, ub: 2e10 },
```

the `ub` is never read. every reference to `tvlUsdUI` in the repo is `.lb`:

```
src/queries/yield.js:67,165,193,282     tvlUsdUI.lb
src/handlers/triggerAdaptor.js:409      tvlUsdUI.lb
```

the only upper bounds enforced anywhere are `tvlUsdDB.ub` (5e10) at insert time in
`triggerAdaptor.js:134` and `apy.ub`. $31.4b is under $50b, so the row is inserted, and nothing
filters it on the way out.

## enabling the existing ub is not the fix

lido stETH is $24,286,038,515 today against a lido protocol tvl of $24,208,098,312. it is real, and
a $20b absolute cap deletes the largest legitimate pool in the dataset.

what separates the two is relative, not absolute:

```
lido stETH            pool $24.29b / protocol $24.21b  =   1.003
aerodrome WETH-SAND   pool $31.43b / protocol $0.13b   = 242.03
```

over all 16,733 pools whose project matches a protocol slug (0 unmatched):

```
ratio > 1.5    29 pools
ratio > 2      23
ratio > 3      15
ratio > 5       9
ratio > 10      5
ratio > 50      1
```

only the aerodrome pool is past 50x, so a conservative multiplier catches this without touching
anything else. `api.llama.fi/protocols` is already fetched in `exclude.js`
(`getProtocolExclusionsFromApi`), so the denominator costs no extra request.

`excludePools` is at 1,941 hand-maintained entries and `excludedProtocols` at 1,018, which is the
same problem being handled one pool at a time — this one is not in either list.

happy to send a patch against `triggerAdaptor.js` if the approach looks right.

0xshubhs:
Fixes #2950

`/pools` is currently serving `aerodrome-slipstream` `WETH-SAND` on Base at
`tvlUsd: 31,431,600,393`, which is 242x the entire protocol's tvl and 17.11% of the
$183,753,699,148 total across all 17,274 pools. the pool
(`0xa69cdf524d072c366fd050957880552916d29405`) holds 2.147927 WETH and 26.2t of a token
coins.llama.fi has no price for at all, so the real number is about $5,352 of weth plus whatever
the SAND side is worth. it read $121,012 on 08-21, $31.4b on 08-22, and has not updated since
while its sibling pools have.

`boundaries.tvlUsdUI` already has a `ub` of 2e10 for exactly this, and the comment above it says
"max ($20 billion)", but the value is never read anywhere in the repo, every reference to
`tvlUsdUI` is `.lb`. it also cannot be turned on as written: lido steth is $24,286,038,515
today against a lido protocol tvl of $24,208,098,312, so a $20b absolute cap deletes the
largest legitimate pool in the dataset.

what separates the two is the ratio to the pool's own protocol, not the absolute size:

```
lido stETH            pool $24.29b / protocol $24.21b  =   1.003
aerodrome WETH-SAND   pool $31.43b / protocol $0.13b   = 242.03
```

so this adds `boundaries.tvlUsdVsProtocol` and filters on it at insert time in
`triggerAdaptor.js`, next to the existing `tvlUsdDB` bounds.

the denominator comes from `api.llama.fi/protocols`, which `getProtocolExclusionsFromApi`
already fetches. i pulled that fetch into a shared `fetchProtocols` so the new lookup does not
add a request; that is the one behaviour change here beyond the filter, and it also means the
two callers no longer fetch the same payload twice. a rejected fetch clears the cached promise
so the next call retries.

the check is deliberately conservative:

- skipped entirely when the protocol is not on api.llama.fi, or its tvl is under `minProtocolTvl`
  ($1m), so a protocol we do not track and one whose tvl adapter is down never lose pools
- a failed fetch returns null and drops nothing
- `multiplier` is 50, which is far looser than needed

run against every pool `/pools` currently returns:

```
projects checked: 367   skipped (unknown/small protocol tvl): 135
pools dropped: 1

  DROP  aerodrome-slipstream | Base | WETH-SAND
        pool tvlUsd $31,431,600,393  vs protocol $129,865,242  = 242.0x

control lido stETH: pool $24,286,038,515 vs protocol $24,208,098,312 = 1.003x -> kept
```

for context on where the threshold could go, the distribution over the 16,733 pools whose
project matches a protocol slug (0 unmatched) is `>1.5: 29, >2: 23, >3: 15, >5: 9, >10: 5,
>50: 1`. 50 is the setting that changes exactly one pool today; anything tighter is a
one-line change to the boundary if you want it.

also verified: 502 `getProtocolTvl` calls issue 1 http request, an unknown slug returns null,
and a simulated fetch failure logs and drops nothing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pool data quality by filtering out pools with implausibly high TVL compared with their protocol’s total TVL.
  * Added safeguards to apply this validation only when reliable protocol TVL data is available.
  * Improved efficiency and reliability when retrieving protocol TVL information by reusing cached data and handling invalid responses safely.

### DIFF
--- a/src/handlers/triggerAdaptor.js
+++ b/src/handlers/triggerAdaptor.js
@@ -134,6 +134,21 @@ const main = async (body) => {
       p.tvlUsd <= exclude.boundaries.tvlUsdDB.ub
   );
 
+  // Filter pools holding implausibly more tvl than the protocol they belong to.
+  const protocolTvl = await exclude.getProtocolTvl(body.adaptor);
+  const { multiplier, minProtocolTvl } = exclude.boundaries.tvlUsdVsProtocol;
+  if (Number.isFinite(protocolTvl) && protocolTvl >= minProtocolTvl) {
+    const ceiling = protocolTvl * multiplier;
+    const implausible = data.filter((p) => p.tvlUsd > ceiling);
+    if (implausible.length) {
+      console.log(
+        `${body.adaptor}: dropping ${implausible.length} pool(s) over ${multiplier}x the protocol tvl of ${protocolTvl}: ` +
+          implausible.map((p) => `${p.symbol} ${p.tvlUsd}`).join(', ')
+      );
+      data = data.filter((p) => p.tvlUsd <= ceiling);
+    }
+  }
+
   // nullify NaN, undefined or Infinity apy values
   data = data.map((p) => ({
     ...p,
--- a/src/utils/exclude.js
+++ b/src/utils/exclude.js
@@ -2495,22 +2495,37 @@ const boundaries = {
   apy: { lb: 0, ub: 1e6 },
   // reading from database returns only pools which is max 7 days old
   age: 7,
+  // a pool cannot hold this many times its own protocol's tvl, only checked when
+  // that protocol is on api.llama.fi with a tvl of at least minProtocolTvl
+  tvlUsdVsProtocol: { multiplier: 50, minProtocolTvl: 1e6 },
+};
+
+let protocolsPromise = null;
+
+const fetchProtocols = async () => {
+  if (!protocolsPromise) {
+    protocolsPromise = (async () => {
+      const response = await fetch('https://api.llama.fi/protocols', {
+        signal: AbortSignal.timeout(10_000),
+      });
+      if (!response.ok) {
+        throw new Error(`HTTP ${response.status}`);
+      }
+      const data = await response.json();
+      return Array.isArray(data) ? data : [];
+    })().catch((err) => {
+      protocolsPromise = null;
+      throw err;
+    });
+  }
+  return protocolsPromise;
 };
 
 const getProtocolExclusionsFromApi = async () => {
   try {
-    const response = await fetch('https://api.llama.fi/protocols', {
-      signal: AbortSignal.timeout(10_000),
-    });
-    if (!response.ok) {
-      throw new Error(`HTTP ${response.status}`);
-    }
-
-    const data = await response.json();
+    const data = await fetchProtocols();
     return new Set(
-      (Array.isArray(data) ? data : [])
-        .filter((p) => p.rugged || p.deprecated || p.deadFrom)
-        .map((p) => p.slug)
+      data.filter((p) => p.rugged || p.deprecated || p.deadFrom).map((p) => p.slug)
     );
   } catch (err) {
     console.log(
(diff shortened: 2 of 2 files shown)


slasher125:
hi 0xshubhs. ty for the analysis #2950. 

I'm closing this in favour of a sibling-based gate (src/utils/tvlDominance.js, ported from our new v2 yield pipeline). it compares a pool against its own adapter batch rather than protocol TVL from /protocols

PR #2956, opened 2026-08-28
Add Phase Delegation (pdSOL) and Phase Stake (YIELD) yield adapters

itsgotto:
Adds yield adapters for Phase's two Solana liquid staking tokens. Both protocols were listed on the TVL side earlier today (#20720 and #20739) and are live under `parent#phase`.

| | project | protocolId | mint | stake pool |
|---|---|---|---|---|
| pdSOL | `phase-delegation` | 8520 | `aeroXvCT6tjGVNyTvZy86tFDwE4sYsKCh7FbNDcrcxF` | `aero2ePURjuEgLKTzcUmF6RypBncBGd7pMUYCoSsVJ6` |
| YIELD | `phase-stake` | 8527 | `phaseZSfPxTDBpiVb96H4XFSD8xHeHxZre5HerehBJG` | `phasejkG1akKgqkLvfWzWY17evnH6mSWznnUspmpyeG` |

**Methodology**

Both follow the existing `jagpool-staked-sol` adapter: `getStakePoolInfo` reads the stake pool account, `calcSolanaLstApy` derives single-epoch APY from the account's own `lastEpochTotalLamports` / `lastEpochPoolTokenSupply` snapshots, and `solanaLstPricePerShare` supplies the exchange rate. TVL is `tvlSol * SOL price`.

Deliberately not using `getSanctumLstApy`, so neither adapter depends on a third-party API or key. Everything is read from chain.

**Verification**

Decoded all three pools with `StakePoolLayout` from `@solana/spl-stake-pool` at epoch 1024:

```
jagpool (control)  prevRate 1.11053617 -> 1.11079950   apy 3.83%
pdSOL              prevRate 1.14902804 -> 1.14929753   apy 3.79%
YIELD              prevRate 1.18482057 -> 1.18518416   apy 4.99%
```

The control reproduces jagSOL's currently published 3.83% exactly, which is what gave me confidence the same path is correct for these two.

Note that YIELD's pool runs on the Sanctum multi-validator stake pool program (`SPMBzsVUuoHA4Jm6KunbsotaahvVikZs1JyTW6iJvbn`, the same program jupSOL uses) rather than the standard SPL program. `StakePoolLayout` decodes it identically, 611-byte account, and the snapshot fields are populated.

TVL at time of writing: pdSOL ~$114.5M, YIELD ~$1.9M.

## Summary by CodeRabbit

* **New Features**
  * Added yield tracking for Phase’s Solana stake and delegation products.
  * Displays APY, total value locked, token pricing, and applicable epoch fees.
  * Includes links to the Phase staking and delegation pages.
  * Added validation to ensure reliable SOL pricing data before displaying metrics.

### DIFF
--- a/src/adaptors/phase-delegation/index.js
+++ b/src/adaptors/phase-delegation/index.js
@@ -1,0 +1,52 @@
+const axios = require('axios');
+const {
+  getStakePoolInfo,
+  calcSolanaLstApy,
+  solanaLstPricePerShare,
+  getPriceApiUrl,
+} = require('../utils');
+
+const PDSOL_MINT = 'aeroXvCT6tjGVNyTvZy86tFDwE4sYsKCh7FbNDcrcxF';
+const STAKE_POOL = 'aero2ePURjuEgLKTzcUmF6RypBncBGd7pMUYCoSsVJ6';
+const SOL = 'So11111111111111111111111111111111111111112';
+
+const solKey = `solana:${SOL}`;
+
+const apy = async () => {
+  const [stakePool, priceRes] = await Promise.all([
+    getStakePoolInfo(STAKE_POOL),
+    axios.get(getPriceApiUrl(`/prices/current/${solKey}`)),
+  ]);
+
+  const solPrice = priceRes.data.coins[solKey]?.price;
+  if (!solPrice) throw new Error('Unable to fetch SOL price');
+
+  const apyBase = calcSolanaLstApy(stakePool);
+  const pricePerShare = solanaLstPricePerShare(stakePool);
+
+  const feePct = stakePool.epochFee
+    ? `${((stakePool.epochFee.numerator / stakePool.epochFee.denominator) * 100).toFixed(0)}% epoch fee`
+    : null;
+
+  return [
+    {
+      pool: PDSOL_MINT,
+      chain: 'Solana',
+      project: 'phase-delegation',
+      symbol: 'pdSOL',
+      tvlUsd: stakePool.tvlSol * solPrice,
+      apyBase,
+      ...(pricePerShare > 0 && { pricePerShare }),
+      underlyingTokens: [SOL],
+      poolMeta: feePct,
+      isIntrinsicSource: true,
+    },
+  ];
+};
+
+module.exports = {
+  protocolId: '8520',
+  timetravel: false,
+  apy,
+  url: 'https://phase.cc/delegation',
+};
--- a/src/adaptors/phase-delegation/index.js
+++ b/src/adaptors/phase-delegation/index.js
@@ -18,8 +18,9 @@ const apy = async () => {
     axios.get(getPriceApiUrl(`/prices/current/${solKey}`)),
   ]);
 
-  const solPrice = priceRes.data.coins[solKey]?.price;
-  if (!solPrice) throw new Error('Unable to fetch SOL price');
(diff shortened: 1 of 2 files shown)


itsgotto:
Thanks, fixed the two price-guard findings in both adapters.

On the stale snapshot point: both of these pools update every epoch, `lastUpdateEpoch` is 1024 and the current epoch is 1024, so neither is stale today. But the underlying concern is valid, and it lives in `calcSolanaLstApy` in `src/adaptors/utils.js` rather than in these adapters, since `jagpool-staked-sol` uses the same path. Happy to handle it however you prefer, though diverging in one adapter seemed worse than leaving the shared helper consistent.

Related, and possibly worth a separate look: `getStakePoolInfo` derives `epochsPerYear` as `currentEpoch / yearsSinceGenesis`, which currently gives ~158.7. Measured against real slot times on mainnet right now, epochs are running ~1.6 days, which is ~228 per year:

```
last   2,000 slots  320.5 ms/slot
last  10,000 slots  320.9 ms/slot
last  50,000 slots  348.5 ms/slot
last 200,000 slots  361.4 ms/slot
```

The lifetime average is dragged down by Solana's earlier, slower epochs, so annualising a single epoch's growth by 158.7 understates APY by roughly 40% for every adapter on this path. For pdSOL it is the difference between the 3.79% this PR currently reports and ~5.5%. Deriving `epochsPerYear` from current epoch duration would fix it across all of them.

I did not want to change shared utils in a listing PR, so flagging rather than patching. Happy to open a separate PR for it if useful.

itsgotto:
One thing worth flagging: the CI run only exercised `phase-delegation`. The log shows `Running phase-delegation Test` and no equivalent for `phase-stake`, so the second adapter has not produced output in your pipeline even though both files are in the diff.

It is almost certainly fine, since the two files are structurally identical and only differ in addresses, but I would rather you knew than have it merge untested. Verified locally against the same `StakePoolLayout` decode the helper uses, at epoch 1024:

```
pdSOL  prevRate 1.14902804 -> 1.14929753   apyBase 3.79%   tvl ~$114.9M
YIELD  prevRate 1.18482057 -> 1.18518416   apyBase 4.99%   tvl ~$1.9M
```

For reference, the same method reproduces `jagpool-staked-sol` at 3.83%, matching its currently published figure, which is what gave me confidence in the approach.

Note that YIELD's pool runs on the Sanctum multi-validator stake pool program (`SPMBzsVUuoHA4Jm6KunbsotaahvVikZs1JyTW6iJvbn`) rather than the standard SPL program. `StakePoolLayout` decodes it identically and the last-epoch snapshot fields are populated, but that is the part most worth a second pair of eyes if you want to trigger a run against it.

itsgotto:
Happy to get this out of your queue.

The adapters are ready as they stand. Both price-guard findings are fixed, CI is green, and the diff touches only the two new adapter files.

The `epochsPerYear` point I raised is a separate concern in `src/adaptors/utils.js` and affects every Solana LST on the `calcSolanaLstApy` path, not just these two. I'll move it into its own issue with the measurements so it isn't sitting on an adapter add.

Not a blocker from my side: the CI log only ran `phase-delegation`, so `phase-stake` has no pipeline output. The two files are structurally identical and I verified the second locally against the same decode path, so I'm happy for it to merge as is if you are.

0xkr3p (approved):
hi @itsgotto, thanks for the PRs, for the future its 1 adapter per PR - approved

PR #2959, opened 2026-08-29
feat: add 3F yield adapter

yieldhunting:
Adds a yields adapter for **3F** (slug `3f`, listed on the TVL page since May, `three-f/index.js`).

3F gives leveraged exposure to tokenized RWAs by looping them through Morpho Blue markets, bridging the RWA's asynchronous settlement window with short-term facilitator loans. Each leverage tier is a separate `PositionManager` — an ERC-20 share token whose `totalAssets()` is that tier's NAV, denominated in the market's debt asset.

Today 3F's positions surface on the yields page only as `morpho-blue` collateral rows (`WJAAA`, `WFALCONX`) with **apy 0**, which is correct from Morpho's side — collateral doesn't earn there — but it means the protocol's actual product, the levered RWA yield, isn't represented anywhere.

### Output

13 pools, one per funded leverage tier:

```
symbol      meta                    tvlUsd   apyBase     ltv
WFALCONX    5x leverage         2517924.47     0.000  0.8039
WFALCONX    6x leverage         1012674.15     0.000  0.8366
WFALCONX    4x leverage          901777.57     0.000  0.7549
WFALCONX    3x leverage          697779.80     0.000  0.6732
WJAAA       8x leverage          428016.76    16.295  0.8775
WFALCONX    7x leverage           97964.52     0.000  0.8492
WJAAA       10x leverage          90256.85    17.352  0.9020
WJAAA       4x leverage            7395.86     9.768  0.7549
WJAAA       7x leverage            5050.08    17.058  0.8599
WJAAA       5x leverage            1336.29    11.627  0.8039
WJAAA       2x leverage            1059.88     7.094  0.5098
WJAAA       1x leverage            1047.52     5.668  0.0000
WJAAA       3x leverage            1037.60     8.385  0.6732
```

The wJAAA ladder is a useful sanity check: the 1x tier reports 5.67%, which is where the underlying Janus Henderson AAA CLO fund yields, and the rate scales monotonically with leverage from there.

Markets covered: wJAAA (Janus Henderson Anemoy AAA CLO), wUSCC (Bitwise Crypto Carry), wFalconX (Pareto AA tranche of the FalconX credit vault) and wmGLO (Midas Fasanara Global Open). Tiers are discovered from `PositionManagerCreated` logs, so new markets and tiers appear without a code change.

### Methodology notes

**`tvlUsd` is depositor equity, not gross collateral.** `totalAssets()` is levered collateral minus debt. This deliberately avoids double-counting the collateral already counted in the underlying Morpho Blue market — the sum here is ~$5.76M against ~$29.2M of gross collateral.

**`apyBase` uses a 7-day realised window, not 24h.** Tokenized RWA funds mark to a NAV oracle that steps once a day or slower, so a 24h window either catches a whole step or none of it. At 8–17x leverage a caught step annualises into the hundreds of percent and a missed one reports zero — the first draft of this adapter produced 182% on the wJAAA 8x tier for exactly that reason. Seven days spans several NAV steps and gives a stable rate. Because it is realised share-price growth, it is already net of borrow cost and of both the management and performance fees (fees are minted as diluting shares), so it is the lower-bound attainable number rather than a modelled `L·r_collateral − (L−1)·r_borrow` spread.

**Why the wFalconX tiers report 0.** Their share price has *declined* over the window — the Pareto AA coupon is currently below the Morpho USDC borrow rate, so the levered carry is inverted and NAV per share is falling. No pool in the yields API currently reports a negative `apyBase`, so these report 0 rather than a negative rate. Flagging it explicitly because it's a live condition on the largest tiers, not a bug in the adapter; they'll report a positive rate once the spread turns.

**Dust tiers are filtered** at `utils.MIN_TVL_USD`. Several deployed tiers hold under a dollar, where annualising a share-price wobble produces meaningless rates.

### Testing

Run directly against mainnet (13 pools, ~5s). Note that `npm run test --adapter=3f` currently fails on `master` for an unrelated reason — a missing Babel plugin in the repo's jest setup, which reproduces on other adapters too:

```
Error: Jest: Got error running globalSetup - src/adaptors/beforeTests.js,
reason: Cannot find module '@babel/plugin-transform-runtime'
```

It fails in `globalSetup` before any adapter is selected, so it is not specific to this PR — `npm run test --adapter=aave-v3` fails identically on a clean checkout. Verification was therefore done by invoking the exported `apy()` directly against mainnet. Happy to add a `@babel/plugin-transform-runtime` devDependency in a separate PR if that would be useful.

## Summary by CodeRabbit

* **New Features**
  * Added support for Ethereum 3F (Grunt) yield pools.
  * Displays leverage tiers with current TVL and annualized APY based on seven-day performance.
  * Excludes pools with insufficient TVL, invalid metadata, or unavailable pricing.
  * Caps reported APY at 100% and shows zero when performance data is unreliable.

### DIFF
--- a/src/adaptors/3f/index.js
+++ b/src/adaptors/3f/index.js
@@ -242,4 +242,7 @@ module.exports = {
   timetravel: false,
   apy,
   url: URL,
+  // DefiLlama protocol id for slug "3f" (defillama.com/protocol/3f,
+  // TVL adapter projects/three-f/index.js).
+  protocolId: '7923',
 };
--- a/src/adaptors/3f/index.js
+++ b/src/adaptors/3f/index.js
@@ -1,0 +1,245 @@
+// 3F (Grunt) — DefiLlama Yield Adapter
+// ===========================================================
+// 3F provides leveraged exposure to tokenized real-world assets. A user picks an
+// RWA vault and a leverage tier; the protocol loops it through a Morpho Blue
+// market behind the scenes, bridging the RWA's asynchronous settlement window
+// with short-term facilitator loans.
+//
+// Each leverage tier is a separate `PositionManager` contract — an ERC-20 share
+// token whose `totalAssets()` is the tier's net asset value, denominated in the
+// market's debt asset (USDC). Levered collateral value minus debt, so it is
+// already net of borrowings and does not double-count the underlying Morpho
+// market's collateral.
+//
+// One pool is published per deployed leverage tier. As of writing that is the
+// wJAAA (Janus Henderson Anemoy AAA CLO), wUSCC (Bitwise USCC) and wFalconX
+// (Pareto AA tranche of the FalconX credit vault) markets.
+//
+// Everything is read on-chain at adapter run-time (no 3F API dependency):
+//   - discovery : PositionManagerCreated logs from the PositionManager factory,
+//                 which also carry each tier's target LTV.
+//   - tvlUsd    : PositionManager.totalAssets() (debt-asset units) x debt-asset
+//                 price from coins.llama.fi. This is depositor equity, not the
+//                 gross levered collateral.
+//   - apyBase   : realised growth in share price (totalAssets / totalSupply)
+//                 over a trailing 7-day window, annualised. This is what a
+//                 depositor in that tier actually earned — already net of
+//                 borrow cost, management fee and performance fee, since fees
+//                 are minted as diluting shares. No projection, no off-chain
+//                 rate, no modelled spread.
+//   - ltv       : the tier's target LTV, straight from the creation event.
+//
+// Why realised share price rather than a modelled spread: the headline yield on
+// a levered RWA position is L*r_collateral - (L-1)*r_borrow, but r_collateral
+// for an async-settling RWA fund is only observable through its NAV oracle, and
+// the borrow leg floats with Morpho utilisation. Share-price growth captures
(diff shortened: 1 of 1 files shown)


yieldhunting:
CI's `protocolId` failures are fixed in `83d5ce2` — all checks are green now (102/102, 12 pools exported).

One note in case the stale CodeRabbit comment above is read as still-outstanding: it suggests `protocolId: '3f'`, the slug. The test wants the numeric protocol id:

```js
expect(protocolId).toBe(String(protocol.id));
```

and the source-literal check only matches digits:

```js
/\bmodule\.exports\s*=\s*\{[\s\S]*?\bprotocolId\s*:\s*['"](\d+)['"]/
```

so `'3f'` would fail both. This PR uses `protocolId: '7923'`, which is 3F's id in `api.llama.fi/config` (slug `3f`, TVL adapter `three-f/index.js`), matching the convention in neighbouring adapters.

Also worth flagging that the pool count moves between 12 and 13 across runs. Several wJAAA tiers sit within a few dollars of `utils.MIN_TVL_USD`, so one crosses the filter boundary between samples. Both counts are correct at their sample time, and DefiLlama's own display threshold is $10k, so those tiers stay hidden either way — happy to raise the adapter's floor if you'd prefer a stable count.

0xkr3p (approved):
thanks for the pr @yieldhunting approved

PR #2960, opened 2026-08-30
Discover Neverland markets from the on-chain registry

0xCatalyst:
Discovers [Neverland](https://defillama.com/yields?protocol=Neverland) lending markets from `PoolAddressesProviderRegistry` instead of maintaining provider addresses in the adapter. Registered markets now receive their metadata, application route, and borrow-routing group from on-chain configuration, so adding another market no longer requires a `yield-server` maintenance.

### Market discovery

`getMarkets` loads `getAddressesProvidersList`, then resolves each provider's `getMarketId` and `getPoolDataProvider`. Invalid entries are ignored and each provider's reserve reads remain isolated, while the adapter rejects a completely empty lending result rather than publishing one silently.

### Pool metadata

`poolMeta` now comes from `getMarketId`. Canonical reserves link to `/markets?asset=<symbol>`, isolated market IDs derive `/isolated/<market>` automatically, and `routeGroupKey` remains the data provider address so borrow routing stays within a market.

Expired streams are still filtered from APY and `rewardTokens` through the same `liveEmissions` timestamp, preserving consistent reward metadata.

## Summary by CodeRabbit

* **New Features**
  * Markets are now discovered automatically from the live protocol registry.
  * Market links and metadata are generated dynamically for isolated markets.
* **Bug Fixes**
  * Improved handling of incomplete reward and pricing data to prevent misleading APY results.
  * Unavailable reward data no longer appears as zero-value APY.
* **Reliability**
  * Clear errors are now reported when market or lending-pool data cannot be resolved.
  * Borrow APY availability is detected more accurately.

### DIFF
--- a/src/adaptors/neverland/abi.js
+++ b/src/adaptors/neverland/abi.js
@@ -1,4 +1,29 @@
 module.exports = {
+  poolAddressesProviderRegistryAbi: [
+    {
+      inputs: [],
+      name: 'getAddressesProvidersList',
+      outputs: [{ internalType: 'address[]', name: '', type: 'address[]' }],
+      stateMutability: 'view',
+      type: 'function',
+    },
+  ],
+  poolAddressesProviderAbi: [
+    {
+      inputs: [],
+      name: 'getMarketId',
+      outputs: [{ internalType: 'string', name: '', type: 'string' }],
+      stateMutability: 'view',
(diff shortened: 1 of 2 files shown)


0xkr3p (approved):
gm @0xCatalyst, thanks for the pr - approved

PR #2961, opened 2026-08-30
feat: add Anemoy Capital yield adapter

yieldhunting:
Adds a yields adapter for **Anemoy Capital** (slug `anemoy-capital`, listed for TVL, adapter `projects/anemoy-capital/index.js`).

Anemoy issues **JTRSY**, the Janus Henderson Treasury Fund — a tokenized fund holding short-dated US Treasuries, distributed through Centrifuge's ERC-7540 asynchronous vaults. It's ~$871M of TVL with no yields coverage today.

### Output

```
chain       symbol              tvlUsd    apyBase
Ethereum    JTRSY            870683978      3.490
```

TVL resolves to $870.68M against the TVL adapter's $870.70M — same valuation path, so the two agree.

### Methodology

JTRSY is a fixed-balance ERC-20; it does not rebase. Yield accrues through the fund's NAV, exposed as `pricePerShare()` on the vault that the LTF contract returns for a given settlement asset (USDC). `tvlUsd` is `totalSupply() × pricePerShare`, priced at the USDC rate from `coins.llama.fi` — deliberately the same path `projects/anemoy-capital/index.js` uses to value the token, so TVL can't drift between the two adapters.

`apyBase` is realised `pricePerShare` growth over a trailing 7-day window, annualised. It's the fund's actual accrual, not a projection or an off-chain rate. The result is stable across window lengths, which is the sanity check that matters for a treasury fund:

| window | annualised |
|---|---|
| 7d | 3.49% |
| 30d | 3.48% |
| 90d | 3.60% |
| 180d | 3.43% |

That's the right level for short-duration US Treasuries.

### Notes

**NAV is read once on Ethereum and applied to every chain's float**, mirroring `getNav()` in the TVL adapter, which does the same. It's a single fund-level number — the per-chain deployments differ only in how much of the float sits there.

**Only Ethereum publishes a pool.** Base and Celo carry the LTF contracts but currently hold zero JTRSY supply, so they're filtered rather than emitted as empty pools. If float moves to those chains they'll appear automatically with no code change.

**Guard:** if the archive read for the prior window fails, or the window shows no growth, the pool publishes `tvlUsd` with no `apyBase` rather than failing the adapter or asserting a rate it can't evidence.

## Summary by CodeRabbit

* **New Features**
  * Added yield tracking for Anemoy Capital’s JTRSY tokenized treasury fund.
  * Reports total value locked and base APY across Ethereum, Base, and Celo.
  * Calculates APY from trailing seven-day NAV performance.
  * Excludes invalid market data and supply values from results.
  * Returns no pools when NAV data is unavailable or shows no valid growth.

### DIFF
--- a/src/adaptors/anemoy-capital/index.js
+++ b/src/adaptors/anemoy-capital/index.js
@@ -1,0 +1,142 @@
+// Anemoy Capital — DefiLlama Yield Adapter
+// ===========================================================
+// Anemoy issues JTRSY, the Janus Henderson Treasury Fund: a tokenized fund
+// holding short-dated US Treasuries, distributed through Centrifuge's
+// ERC-7540 asynchronous vaults on Ethereum, Base and Celo.
+//
+// JTRSY is a fixed-balance ERC-20 — it does not rebase. Yield accrues through
+// the fund's NAV, exposed as `pricePerShare()` on the vault that the LTF
+// contract returns for a given settlement asset (USDC here). This mirrors how
+// the TVL adapter (projects/anemoy-capital) values the same token.
+//
+//   - tvlUsd  : LTF.totalSupply() x pricePerShare, both 6-decimal, USDC-
+//               denominated, then priced at the USDC rate from coins.llama.fi.
+//   - apyBase : realised growth in pricePerShare over a trailing 7-day window,
+//               annualised. This is the fund's actual accrual — no projection
+//               and no off-chain rate.
+//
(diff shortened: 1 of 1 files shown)


0xkr3p:
hey @yieldhunting, thanks for the PR! Sorry but we won't be moving forward as this is a duplicate pool. we currently track under centrifuge: https://defillama.com/yields/pool/ff1bb959-d160-4906-bad2-d3e37a1e92e8

PR #2704, opened 2026-05-26
Satsuma: index all Citrea LP pools and ICHI vaults

MiguelBits:
## Summary

- **satsuma** adapter: bump analytics subgraph to v1.0.3 and filter pools with TVL > 0 so all live Citrea LP pairs are indexed (including WCBTC/USDC.e).
- **ichi** adapter: expand Citrea vault allowlist from 3 to 16 addresses (all Satsuma ICHI vaults on Citrea).

## Test plan

- [ ] \
pm run test -- --adapter=satsuma\
- [ ] \
pm run test -- --adapter=ichi\

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Chores**
  * Expanded vault data source configuration for Citrea network
  * Updated Satsuma data source to latest version

* **Bug Fixes**
  * Improved APY data filtering to ensure only valid pools are included

[Review Change Stack](https://app.coderabbit.ai/change-stack/DefiLlama/yield-server/pull/2704?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #2969, opened 2026-09-02
fix(satsuma): replace deleted subgraph with on-chain data source

tiagotpratas:
Resubmission of #2966 with a clean diff (that PR was opened from a non-fork repo and carried the whole tree; closed by me).

Satsuma's adapter queries Goldsky `analytics-mainnet/v1.0.3`, which was **deleted upstream**. Every request 404s, so Satsuma has been silently missing from yields despite the adapter being merged in #2704.

This rebuilds it on sources that can't go stale the same way:

- **TVL and 24h volume** from Satsuma's own public `/pool-stats` endpoint (built by scanning Swap logs over RPC).
- **token0/token1, symbols and the pool fee** read directly from the pools via batched `eth_call` on Citrea mainnet.

Algebra pools use a **dynamic fee**, so the fee is read from `fee()` per pool rather than assumed. Live values range from 0.01% to 1.5% across Satsuma pools.

Total TVL agrees with the existing DefiLlama Satsuma TVL listing (~$780k). 14 pools exported.

https://claude.ai/code/session_01Szi7Q56pDzLiweiCZKrhZV

## Summary by CodeRabbit

- **New Features**
  - Satsuma pool statistics now include TVL and 24-hour trading volume.
  - Pool details now use on-chain data for token addresses, fees, and symbols.
  - APY is calculated from 24-hour volume and fee rates.
  - Pool results now include eligible gauge reward APY and reward token information.
  - SUMA pricing is derived from pool data and coin pricing, with a parity fallback.

- **Updates**
  - The 7-day APY metric is no longer included in Satsuma pool information.

### DIFF
--- a/src/adaptors/satsuma/index.js
+++ b/src/adaptors/satsuma/index.js
@@ -8,6 +8,11 @@ const CHAIN = 'citrea';
 // Satsuma's analytics subgraph was deleted upstream, which silently dropped
 // every Satsuma pool from this adapter. Both sources below are on-chain or
 // derived from on-chain logs, so there is no indexer to go stale again.
+//
+// /pool-stats returns, per Algebra pool: tvlUSD, volumeUSD24h, feesUSD24h,
+// token0/token1 + decimals, and a `prices` map (USD per token) that is the
+// same one used to compute tvlUSD, so vault holdings valued with it are
+// consistent with the pool TVL.
 const POOL_STATS = 'https://s33-epoch-cron.tiagopratas69.workers.dev/pool-stats';
 const RPC = 'https://rpc.mainnet.citrea.xyz';
 // Axios defaults to no timeout; bound every upstream call so apy() rejects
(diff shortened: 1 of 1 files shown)


tiagotpratas:
Pushed a second commit adding reward APY from the s33 gauge emissions (weekly xSATS per pool, SUMA priced on-chain from the SUMA/ctUSD pool). Adapter is fully on-chain plus one public stats endpoint; no subgraph.

tiagotpratas:
Addressed all three review comments in the latest commit: 30s timeout on every axios call (RPC, stats endpoint, coins API), `ethCallBatch` now throws on any JSON-RPC item with `error` or no `result`, and `apyBase` uses the endpoint's `feesUSD24h` (accumulated per swap at the fee in force) instead of `volume * fee()`. The `fee()` call is gone. Verified against Citrea mainnet: 14 pools, same APY values.

tiagotpratas:
Agreed, fixed in the latest commit. Emissions are no longer published on the AMM pool records. Each live IchiVaultGauge now gets its own record (`poolMeta: 'ICHI vault'`, `pool: <gauge>-citrea`) whose TVL is the combined `getTotalAmounts()` of the gauge's two ICHI vaults, valued with the same price map the stats endpoint uses for pool TVL (now exposed as `prices`). `apyReward` is computed against that vault TVL; `apyBase` carries the pool's fee rate since the vaults hold positions in the pool. Verified on Citrea: 14 pool records + 10 vault records.

0xkr3p (commented):
Thanks for this. Moving off the deleted subgraph is the right call, and a lot of it is already solid: pool IDs match the 12 already registered, so no history is lost; each pool has its own url; apyBase uses a 24h window; ethCallBatch fails loudly on RPC item errors; and the dynamic Algebra fee uses the actual feesUSD24  rather than volume × current fee

A few things need to change before merge:

  1. TVL is double-counted
  
The ICHI vaults are LPs inside the AMM pools, so their liquidity is already counted in the AMM rows. For example, the CTR-CTUSD vault holds $107.8k of that pool's $112.1k. AMM-only comes within ~5% of the protocol page.

 Keeping separate rows is right, since only gauge depositors earn emissions. But net them so they partition the pool: `ammTvlUsd = Math.max(0, poolTvlUsd - totalVaultTvlUsdForPool)`

  2. xSATS rewards need the exit penalty applied

  The reward token 0x732bcf02…05ac is xSATS
  (symbol veSUMA). The verified source shows:

  - exit() returns SUMA  immediately but burns 50%  (SLASHING_PENALTY = 5000 / BASIS = 10_000)
  - createVest() returns 100%  only after 180 days,  linearly from 50%
  - Transfers are whitelist-only

  DefiLlama publishes the minimum attainable yield, meaning the post-slashing value when early exit is penalised. apyReward should therefore value xSATS at 50% of SUMA:
  `const rewardUsdPerYear =
  g.weeklySuma * sumaPrice *
  0.5 * 52;`

  Radiant and Geist handle their 50% exit penalty the same way.

 3. Don't use utils.formatSymbol

 Pass symbol as a raw string  (`${symbol0}-${symbol1}`). The server formats symbols
 itself.

  4. Data source

 s33-epoch-cron.tiagopratas69.workers.dev supplies TVL, fees, volume and every
 price. The PR describes it as "Satsuma's own public endpoint" and says "there is no indexer to go stale again". But it's a personal workers.dev subdomain, and the adapter is breaking right now because a single hosted data source went away.

 TVL doesn't need it. I checked the three largest pools: worker TVL exactly equals on-chain balanceOf * price, and coins.llama.fi prices all of their tokens. Could you:
  - compute TVL on-chain with  DefiLlama prices

Thanks again for the updates on the PR

tiagotpratas:
Thanks for the thorough review, all addressed in 6febc0d:

- **TVL double count:** AMM rows now net out the ICHI vaults' in-pool liquidity, so AMM + vault rows partition each pool.
- **xSATS:** valued at 50% of SUMA (instant-exit slash).
- **formatSymbol:** removed, raw `${symbol0}-${symbol1}`.
- **Data source:** the workers.dev endpoint is gone. TVL is on-chain `balanceOf` priced with coins.llama.fi. SUMA and ZNT aren't listed on coins.llama.fi, so those two are priced from the Satsuma pool pairing them with a priced token (only if that side holds ≥ $1k). Happy to switch if you'd rather add them to the coins server.
- **24h fees:** there's no Satsuma-hosted endpoint, so fees now come straight from the pools' Swap logs. Each swap is preceded by a pool event carrying the fee actually applied, which makes the dynamic fee exact per swap. A correction to my earlier description: the old endpoint used volume × current fee, not realised fees. I also now subtract the pool's 50% community fee, so apyBase is what LPs actually receive (the previous version overstated it ~2×).

0xkr3p (approved):
thanks for resolving quickly @tiagotpratas, approved

PR #2977, opened 2026-09-04
Add Lodestar Protocol (Flare) - lodUSDT0 lending pool

lodestarprotocol:
Adds the yield adapter for **Lodestar Protocol** on Flare (protocolId `8565`, listed via DefiLlama-Adapters PR #20759).

Lodestar is no-liquidation fixed-term lending. Lenders supply USD₮0 to an ERC-4626 vault (`lodUSDT0`) and earn the one-time fees borrowers pay up front. There are no emissions, so the only yield is the share price rising.

**How the APY is computed:** realised share-price change over a trailing 30-day window (or since genesis, whichever is shorter), annualised — i.e. exactly what a depositor would have earned. It reads 0 while the book has no loans rather than quoting a projection.

- Pool: `0x87b09bE7A253C2af187c9af17cDEDcEAf4A9780E` (LodestarPool, ERC-4626)
- Asset: USD₮0 `0xe7cd86e13AC4309349F30B3435a9d337750fC82D` (6 dp)
- Reports `tvlUsd`, `apyBase`, `totalSupplyUsd`, `totalBorrowUsd`, `ltv`

`npm run test --adapter=lodestar-protocol` passes (15/15).

Note: this is **not** the same protocol as `lodestar` / `lodestar-v1` (Lodestar Finance on Arbitrum). The slug `lodestar-protocol` follows the project's domain, lodestarprotocol.xyz, to keep the two apart.

## Summary by CodeRabbit

* **New Features**
  * Added Lodestar Protocol support on Flare.
  * Added lending pool metrics, including TVL, borrows, supply, loan-to-value, and annualized base APY.
  * Annualized base APY now reflects realized share-price changes over a trailing 30-day period.
  * Added USDT0 pricing with a $1 fallback when market pricing is unavailable.
  * Added safeguards for newer pools and unavailable historical data, defaulting APY to 0 when reliable calculations are unavailable.

### DIFF
--- a/src/adaptors/lodestar-protocol/index.js
+++ b/src/adaptors/lodestar-protocol/index.js
@@ -1,8 +1,7 @@
 // Lodestar (Flare): no-liquidation, fixed-term lending. Lenders supply USDT0 to an ERC-4626 pool
 // (lodUSDT0) and earn the one-time fees borrowers pay up front; nothing is emitted, so the only yield
 // is the share price rising. The APY below is therefore the realised share-price change, annualised
-// over a trailing window, which is exactly what a depositor would have earned. It reads 0 while the
-// book is empty rather than quoting a projection.
+// over a trailing window, which is exactly what a depositor would have earned.
 //
 // NOT the same protocol as `lodestar` / `lodestar-v1` (Lodestar Finance on Arbitrum). Slug
 // `lodestar-protocol`, after lodestarprotocol.xyz.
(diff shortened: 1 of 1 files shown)


lodestarprotocol:
Thanks for the review — four of the five findings were real and are fixed in `3158fc7`. I verified each against the deployed contract rather than taking them on faith, so noting what I confirmed and the one I did not apply.

**Fixed.**

1. **Genesis timestamp.** Correct, `1756425600` is 2025-08-29. Now `1787961600`. This one mattered more than it looks: with the wrong constant the window clamped to genesis while `days` stayed near 30, so a 6-day return was being annualised as if it were 30 days.

2. **TVL double-count.** Confirmed from source. `LodestarPool.totalAssets()` returns `bal + principalOut - impairedLoss + credit - unvestedFee` — it already includes outstanding principal. Using it for `tvlUsd` and then adding `totalBorrowUsd` for `totalSupplyUsd` counted every borrowed dollar twice. Now `totalSupplyUsd = totalAssets`, `totalBorrowUsd = principalOut`, `tvlUsd = supply - borrow`. It reads the same today only because `principalOut` is 0.

3. **Negative APY.** Confirmed the share price can fall: a settlement shortfall past the first-loss buffer is booked to `impairedLoss`, which `totalAssets()` subtracts. `Math.max(0, ...)` would have reported a real lender loss as break-even. Removed, the value is signed now.

4. **Short-window annualisation.** Not one of your points, but the same class and worth closing. Borrowers pay their entire fee up front and it vests into the share price over 7 days, so the price moves in steps. Annualising a 1-day window across one such step produces a triple-digit APY out of a single loan. Below 7 days of history the pool now reports 0 rather than that number.

**Not applied.**

Gating `apyBase` on `principalOut > 0`. The reasoning does not hold for this design: fees vest over 7 days, so after the last loan repays the share price keeps rising for a week and lenders genuinely are still earning during it. Zeroing there would under-report real yield. The 7-day minimum window in (4) covers the actual risk you were pointing at.

One note on the schema: omitting `apyBase` fails validation (`apyFields` must contain a finite number), so "not enough history yet" is expressed as 0.

`npm run test --adapter=lodestar-protocol` — 15/15 passing.

0xkr3p:
hi @lodestarprotocol, thanks for the PR pls ping when the pool is closer to 10k tvl threshold as currently the pool wouldn't be displayed on frontend

PR #2980, opened 2026-09-06
hipo: read the widened treasury state, and drop the second call

behrang:
Hipo upgraded its treasury contract on 2026-09-06. `get_treasury_state` mirrors the contract's storage order, and the upgrade inserted `deficit` at index 5 and `round_duration` + `last_settled_round` after the exchange-rate pair, taking the tuple from 21 values to 24. Every index from 5 on moved, so this adapter was reading `loan_codes` and `previous_rate` where it expected the rate pair.

Indices updated: `previous_rate` 11 → 12, `current_rate` 12 → 13.

## The APY period now comes off the same read

It used to be derived from a second `get_times` call as `next_round_since - current_round_since`, which is a round **length**. `round_duration` is the interval the two rates actually grew over, measured on chain between the last two settled rounds.

The two agree while the pool lends into every round, and diverge exactly when it does not: the treasury only moves the rates when a round it lent into settles, so a skipped round roughly doubles the growth per update while a round length stays put. Annualising by the round length would hold the reported APY steady for a pool whose real rate of growth had halved.

That removes one toncenter request per run, so the 1s `sleep` that spaced the two calls goes too, along with the now-unused helper.

## Verification

Ran the adapter's arithmetic against live contract state:

```
previous_rate=1163036853  current_rate=1163402268  round_duration=65536

apyBase          = 16.318861%
apyBaseInception = 5.444984%
```

`16.318861%` is what the protocol's own gauge and MCP server publish for the same rate pair, so the three agree to six decimal places.

I am from the Hipo team.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hGRAM APY reporting by reading exchange-rate and round-duration data directly from the treasury state.
  - Updated APY data handling to remain compatible with the expanded treasury state introduced in the September 2026 upgrade.
  - Removed an unnecessary delay during APY data retrieval, helping results load more efficiently.

### DIFF
--- a/src/adaptors/hipo/index.js
+++ b/src/adaptors/hipo/index.js
@@ -29,31 +29,26 @@ module.exports = {
       );
     }
 
-    await sleep(1000);
-
-    const getTimes = await utils.getData(
-      'https://toncenter.com/api/v3/runGetMethod',
-      {
-        address,
-        method: 'get_times',
-        stack: [],
-      }
-    );
-    if (getTimes.exit_code !== 0) {
-      throw new Error(
-        'Expected a zero exit code, but got ' + getTimes.exit_code
-      );
-    }
-
     // The treasury stores the hGRAM/GRAM exchange rate before and after the
-    // latest round's loan repayments (fixed-point, 1e9 = 1.0). The rate is
-    // updated once per validation round, so the growth between the two rates
-    // accrued over a single round duration.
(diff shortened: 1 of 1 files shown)


0xkr3p (approved):
hi @behrang, thanks for the change - LGTM

behrang:
Hi @0xkr3p and thank you.

PR #2988, opened 2026-09-08
add stackingdao yields + fix zest v1/v2 adapters

y0s0n:
Adds a stackingdao adapter and fixes both Zest adapters, which were reading retired contracts and missing most of the current markets.

stackingdao (new, protocolId 3934)

- stSTX and stSTXbtc APY are derived on chain the same way the protocol's own API computes them: the latest claim-rewards batch paid into rewards-pox5-v1, projected to a full cycle, net of pool commission, annualized over 25 cycles, split by the rewards contract's bps and divided by each pool's STX value. Output matches the protocol API within 0.02 points.
- stBTC APY is trailing growth of the sbtc-per-stbtc ratio over a 30/14/7 day window at historical tips. The pool is omitted rather than estimated if no window is measurable.
- 36 requests per run, all through Hiro's public API with the shared withRetry helper.

zest-v2

- Reads the current v0-5-data reader instead of v0-1-data, which does not know the stBTC market.
- Adds the stBTC vault, priced as sBTC since the token has no feed and is backed ~1:1.
- Fixes the stSTXbtc price key to the suffixed form the coins server resolves.
- Adds totalSupplyUsd and totalBorrowUsd so the borrow side shows. tvlUsd is supply minus borrows.

zest-v1

- Reads pool-read-v2-1-4 instead of the retired pool-read-v1-3-2.
- Covers all eight priced assets including sBTC, which is most of its TVL, and prices via the coins API instead of Zest oracles.

All three pass the harness. Contract reads go through axios so withRetry can act on 429s, which Hiro's public API returns readily.

## Summary by CodeRabbit

* **New Features**
  * Added yield data for StackingDAO on Stacks, covering stSTX, stSTXbtc, and stBTC pools.
  * Added stBTC market data to Zest v2.
  * Added borrowing metrics, supply and borrowing totals, and loan-to-value data to Zest pool results.

* **Improvements**
  * Updated supported assets across Zest integrations, including sBTC, stSTXbtc, USDh, USDT, ALEX, and USDCx.
  * Improved reliability of on-chain data and asset pricing.
  * Updated Zest’s link to its legacy market page.

### DIFF
--- a/src/adaptors/stackingdao/index.js
+++ b/src/adaptors/stackingdao/index.js
@@ -37,6 +37,7 @@ const TXS_PER_PAGE = 50;
 const MAX_TX_PAGES = 4;
 
 const RETRY = { retries: 3, delayMs: 8000 };
+const HTTP = { timeout: 30000 };
 
 const toNum = (cv) => {
   const v = cvToValue(cv);
@@ -47,16 +48,16 @@ const toNum = (cv) => {
 const readOnly = async (contract, fn, tip) => {
   const [address, name] = contract.split('.');
   const url = `${HIRO}/v2/contracts/call-read/${address}/${name}/${fn}${tip ? `?tip=${tip.replace(/^0x/, '')}` : ''}`;
-  const { data } = await withRetry(() => axios.post(url, { sender: DEPLOYER, arguments: [] }), RETRY);
+  const { data } = await withRetry(() => axios.post(url, { sender: DEPLOYER, arguments: [] }, HTTP), RETRY);
   if (!data.okay) throw new Error(`${contract}.${fn} failed: ${data.cause}`);
   return toNum(hexToCV(data.result));
 };
 
-const getJson = async (path) => (await withRetry(() => axios.get(`${HIRO}${path}`), RETRY)).data;
+const getJson = async (path) => (await withRetry(() => axios.get(`${HIRO}${path}`, HTTP), RETRY)).data;
 
 const fetchPrices = async () => {
   const { data } = await withRetry(
-    () => axios.get(getPriceApiUrl('/prices/current/coingecko:blockstack,coingecko:bitcoin')),
+    () => axios.get(getPriceApiUrl('/prices/current/coingecko:blockstack,coingecko:bitcoin'), HTTP),
     RETRY
   );
   return {
@@ -119,7 +120,7 @@ const fetchLatestClaimBatch = async () => {
 };
 
 const fetchLstApys = async (prices) => {
-  const [batch, pox, commissionBps, ststxbtcBps, ststxBps, supplyBtcV1, supplyBtcV2, supplyStstx, ratio] =
+  const [batch, pox, commissionBps, ststxbtcBps, ststxBps, supplyBtcV1, supplyBtcV2, supplyStstx, liveEscrow, ratio] =
     await Promise.all([
       fetchLatestClaimBatch(),
       getJson('/v2/pox'),
@@ -129,11 +130,12 @@ const fetchLstApys = async (prices) => {
       readOnly(CONTRACTS.ststxbtcTokenV1, 'get-total-supply'),
       readOnly(CONTRACTS.ststxbtcTokenV2, 'get-total-supply'),
       readOnly(CONTRACTS.ststxToken, 'get-total-supply'),
+      readOnly(CONTRACTS.dataStx, 'get-live-escrow'),
       readOnly(CONTRACTS.dataStx, 'get-stx-per-ststx'),
     ]);
 
   const ststxbtcStx = (supplyBtcV1 + supplyBtcV2) / 1e6;
-  const ststxStx = (supplyStstx / 1e6) * (ratio / 1e6);
+  const ststxStx = (Math.max(supplyStstx - liveEscrow, 0) / 1e6) * (ratio / 1e6);
   const result = { ststxStx, ststxbtcStx, ststx: null, ststxbtc: null };
   if (!batch || batch.grossSats <= 0 || ststxStx <= 0 || ststxbtcStx <= 0) return result;
 
@@ -157,7 +159,10 @@ const resolveTipAtBurnHeight = async (burnHeight) => {
   for (let offset = 0; offset <= 10; offset++) {
     const candidates = offset === 0 ? [burnHeight] : [burnHeight + offset, burnHeight - offset];
     for (const h of candidates) {
-      const data = await getJson(`/extended/v2/burn-blocks/${h}/blocks?limit=1`).catch(() => null);
+      const data = await getJson(`/extended/v2/burn-blocks/${h}/blocks?limit=1`).catch((error) => {
+        if (error.response?.status === 404) return null;
+        throw error;
+      });
       const block = data?.results?.[0];
       if (block?.index_block_hash) return block.index_block_hash;
     }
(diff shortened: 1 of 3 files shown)


0xkr3p (commented):
hi @y0s0n, thanks for the PRs -  pls can you split into individual per adapter this makes them easier to review, test and merge faster

0xkr3p:
@y0s0n  Thanks for this. The zest v1/v2 halves look good both reconcile against their protocol pages within ~1.4% (zest-v2 7 pools / $68.3M, zest-v1 8 pools / $2.65M), `tvlUsd = supply − borrow` is right on both, and the per-market isolation is the correct pattern. 

  Three things before merge:

  1. stackingdao: the reward projection freezes at the last claim tx

  The PR says output matches the protocol API within 0.02 points. Against
  `app.stackingdao.com/api/stats` today:

      pool        adapter      protocol API     gap
      stSTX        3.3511%        4.78%      -1.43 pts  (-29.9%)
      stSTXbtc     3.4449%        5.03%      -1.59 pts  (-31.5%)
      stBTC        1.1678%        2.38%      -1.21 pts  (-50.9%)

  I instrumented `fetchLstApys` to find out why, and the projection is the issue:

      batch.cycleId     = 143      <- the current, in-progress cycle
      batch.burnHeight  = 967405   <- height of the LAST CLAIM
      progress          = 0.5275
      multiplier        = 1.8009   <- 0.95 / 0.5275

  `progress` is measured at the burn height of the last claim, not at the current block. That
  matters because the cycle has since finished:

      cycle 143 start burn height   966350
      elapsed at the last claim       1055  -> progress 0.5275   <- what the adapter uses
      elapsed NOW                     2024  -> progress 1.0120
      blocks since the last claim      969  (~6.7 days of bitcoin blocks)

  The reward phase is over, no further claims have landed in almost a week, and the adapter is
  still scaling that cycle's completed payout by 1.8009 as if only half of it had arrived. Had
  the multiplier been computed against current progress it would be 1.0000 — no projection at
  all. I re-ran this four days apart and `batch.burnHeight`, `grossSats`, `progress` and
  `multiplier` were byte-identical both times, so it isn't a transient sample.

  Two ways to fix it, either works:

  - select the highest cycleId that is **not** `pox.current_cycle.id` (the last completed
    cycle), where no projection is needed at all; or
  - compute `progress` from the **current** burn height rather than `batch.burnHeight`, so the
    multiplier converges to 1 as the cycle completes.

  I tested the first by patching the cycle selection to use cycle 142: stSTX came out at
  4.351% and stSTXbtc at 4.355%, i.e. error down from ~1 point to ~0.2-0.3.

  2. The residual gap

  Note the shortfall is ~30% *despite* the 1.8x over-projection, which means the `grossSats`
  the adapter sees for a cycle is materially short of what the protocol books. Fixing the
  multiplier alone won't close it. Could you say what else feeds the figure on your side —
  are there reward sources outside the `claim-rewards` batches into `rewards-pox5-v1`?

  `stBTC` is a separate path (trailing 30/14/7d growth of `get-sbtc-per-stbtc` at historical
  tips) and is ~51% low on its own. A trailing realised measure can legitimately differ from a
  forward projection, but that's a big gap — worth confirming the window is actually spanning
  30 days and not silently falling back to 7.

  3. Smaller items

  - **zest-v2 omits `ltv`** while zest-v1 sets it (0.4 for sBTC, 0.5 for stSTX/STX/stSTXbtc).
    v2 reads `get-asset-apys`, which only returns supply/borrow APY — is the collateral factor
    exposed elsewhere on `v0-5-data`? Without it, v2 won't show in the borrow aggregator.
  - **stackingdao is missing `isIntrinsicSource: true` and `pricePerShare`.** It's category
    Liquid Staking and the adapter already reads `get-stx-per-ststx` and `get-sbtc-per-stbtc`
    internally, so `pricePerShare` is free

y0s0n:
@0xkr3p 

I've addressed all three in 2b3fb1bd:

**1. Reward projection**, fixed. The multiplier was computed from the burn height of the last claim, so once a cycle's claims stopped landing it kept scaling a completed payout. Projection is now by pox-5 distribution coverage instead: the batch covers the highest distribution (half-cycle / end) any of its claims reached, and the multiplier is `2 / covered`, so it converges to 1 once both distributions are claimed.

**2. The residual gap**, the missing pot was the bond managers. The claim scan covered the public signer roster (`signer-manager-<slug>-v1`) but not the six bond managers (`signer-manager-bond-1..6-v2`), which stake through bond events and also claim into `rewards-pox5-v1`, and the contract's split bps are shares of everything that lands there. With their claims included, output now matches `app.stackingdao.com/api/stats` within ~0.02 points on all three pools:

```
pool        adapter    protocol API
stSTX       4.65%      4.66%
stSTXbtc    4.90%      4.92%
stBTC       2.37%      2.36%
```

So to answer the question directly: no sources outside the `claim-rewards` batches, the batch was just being read from an incomplete manager list.

`stBTC` was a separate bug: it was a trailing realized measure (growth of `get-sbtc-per-stbtc` over a 30/14/7d window at historical tips), which is why it read ~51% low against your forward-looking comparison. It now uses the same source as the protocol API: compound growth of the effective ratio (on-chain ratio + stBTC's share of `get-ready-to-release`) from the start of the reward stream (burn 967405, ratio 1.00117674). No more historical-tip reads, which also drops the request count and removes the 429-prone block-scan path.

**3. `ltv` on zest-v2**, added, with one caveat that you need to be aware of: in v2, **LTVs are not per-asset**. They live on efficiency groups, and a position's group (and therefore its LTVs) depends on the combination of collateral assets in it. What the adapter publishes is the group the protocol's own resolver returns for that collateral in isolation, `v0-egroup.resolve(2^assetId)`, read live, so superseded groups are skipped automatically. Current values: STX/stSTX/stSTXbtc 0.4, USDCx 0.5, sBTC/stBTC 0.8. USDh is borrow-only (no egroup covers it), so it gets no `ltv`. A multi-collateral position can resolve to a different group, but that can't be expressed in a per-pool field, the isolated-collateral value is the closest well-defined number that I've added here per your request. 

`isIntrinsicSource: true` and `pricePerShare` are also set on the stackingdao pools now (stSTX: `get-stx-per-ststx`, stBTC: `get-sbtc-per-stbtc`, stSTXbtc: 1, non-rebasing).

y0s0n:
Follow-up on the latest review round:

**Cycle selection (stackingdao):** fixed. `fetchLatestClaimBatch` now walks claim cycles newest to oldest and returns the first one with a usable reward batch (`grossSats > 0` and coverage > 0), so a zero-sat claim or a claim that lands before the first distribution on the newest cycle can no longer suppress both LST pools while an older usable cycle exists.

**Effective-ratio denominator (stackingdao):** fixed. Confirmed in the `data-stbtc-v1` source that `get-sbtc-per-stbtc` divides active backing by `total supply - pending-shares`, so the accrued-but-unreleased addition now uses that same active-share denominator instead of raw total supply. Also confirmed `get-ready-to-release` is a scheduled (pre-commission) amount and the contract takes commission on newly claimed rewards only, so nothing is deducted from it here.

**Pair-specific LTV (zest-v2):** skipped; I believe the premise doesn't hold on this protocol version. The efficiency-group registry on mainnet carries single-bit collateral-only masks (each live egroup's mask is exactly one collateral asset's bit), and the market itself resolves groups with `(get-egroup (pow u2 asset-id))`, collateral only. The debt asset does not enter the mask anywhere in the current contracts, so `v0-egroup.resolve(2^assetId)` is the protocol's own resolution semantics, not an approximation I introduced. If a future version moves to collateral-debt pair groups, the adapter's resolve call would pick that up unchanged since it reads the registry live. 

Re-verified against `app.stackingdao.com/api/stats` after the changes: stSTX 4.64 vs 4.66, stSTXbtc 4.90 vs 4.92, stBTC 2.38 vs 2.36.

PR #2997, opened 2026-09-12
[add] yield-basis - add yield adapter

chadoagent:
Yield Basis (protocol id 6771) is indexed on DefiLlama for TVL ($129.4M) but has no rows at all in yields, so the pools are invisible to anyone comparing BTC yields on the site.

This picks up the adapter from #2718 (opened and closed by the YB team in June) and brings it up to date:

* the market roster is read from the factory (`market_count()` + `markets(i)`) instead of a hardcoded id list. There are 11 markets today, several of them earlier versions of the same asset, and the old list covered only 4 of them.
* markets whose LT reports `is_killed` are skipped. This matches how `fees/yield-basis` in dimension-adapters enumerates the same roster.
* `protocolId: '6771'` added, which the current test suite requires.

Rows emitted, two per market:

* unstaked LT: `apyBase` from `pricePerShare` growth over 24h, `apyBase7d` over 7 days, `pricePerShare` included.
* staked gauge: `apyReward` from YB emissions (`preview_emissions` on the gauge controller, one day apart), `rewardTokens` = YB.

Test run: `npm run test --adapter=yield-basis` -> 169/169 passing. 19 pools, $130.9M in total, against the $129.4M the protocol page reports, so the coverage lines up.

One note on the numbers: `pricePerShare` on these markets carries the cost of rebalancing a 2x leveraged position, so the 24h `apyBase` swings day to day and can be negative. That is the real return of the unstaked side rather than an artefact, and it is why `apyBase7d` is also provided. The fees adapter in dimension-adapters reports the same quantity and carries `allowNegativeValue: true` for the same reason.

## Summary by CodeRabbit

- **New Features**
  - Added support for tracking Yield Basis Ethereum markets.
  - Displays separate staked and unstaked pool records.
  - Includes TVL, share prices, token supplies, symbols, and gauge reward/APY data.
  - Uses historical share-price data for unstaked APY calculations.
  - Pool links direct users to the relevant Yield Basis earn page.
- **Bug Fixes**
  - Improved handling of invalid markets and unavailable pricing, balances, supplies, or reward data.

### DIFF
--- a/src/adaptors/yield-basis/index.js
+++ b/src/adaptors/yield-basis/index.js
@@ -1,0 +1,347 @@
+const sdk = require('@defillama/sdk');
+const axios = require('axios');
+const BigNumber = require('bignumber.js');
+const utils = require('../utils');
+
+const CHAIN = 'ethereum';
+const PROJECT = 'yield-basis';
+const URL = 'https://app.yieldbasis.com/';
+const FACTORY = '0x370a449FeBb9411c95bf897021377fe0B7D100c0';
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
+const DAY = 24 * 60 * 60;
+const DAYS_PER_YEAR = 365;
+const ONE = new BigNumber(1e18);
+
+const ABI = {
+  marketCount: 'uint256:market_count',
+  markets:
+    'function markets(uint256 arg0) view returns ((address asset_token, address cryptopool, address amm, address lt, address price_oracle, address virtual_pool, address staker))',
+  gaugeController: 'address:gauge_controller',
+  token: 'address:TOKEN',
+  previewEmissions:
+    'function preview_emissions(address gauge, uint256 at_time) view returns (uint256)',
+  pricePerShare: 'uint256:pricePerShare',
+  isKilled: 'bool:is_killed',
+  totalSupply: 'erc20:totalSupply',
+  symbol: 'erc20:symbol',
+  updatedBalances:
+    'function updated_balances() view returns (uint256 supply, uint256 staker_balance)',
+};
+
+const isAddress = (address) =>
+  typeof address === 'string' &&
+  /^0x[0-9a-fA-F]{40}$/.test(address) &&
+  address.toLowerCase() !== ZERO_ADDRESS;
+
+const tupleValue = (value, key, index) => value?.[key] ?? value?.[index];
+
+const toAddress = (value) =>
+  isAddress(value) ? value.toLowerCase() : ZERO_ADDRESS;
+
+const toBigNumber = (value) => {
+  try {
+    return new BigNumber(value == null ? 0 : value.toString());
+  } catch (_) {
+    return new BigNumber(0);
+  }
+};
+
+const annualize = (current, previous, days) => {
+  if (!current || !previous || previous.lte(0) || current.lte(0)) return 0;
+  const ratio = current.div(previous).toNumber();
+  if (!Number.isFinite(ratio) || ratio <= 0) return 0;
+  return (Math.pow(ratio, DAYS_PER_YEAR / days) - 1) * 100;
+};
+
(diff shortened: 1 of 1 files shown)


chadoagent:
One question for the reviewers on `apyBase`.

For these markets `pricePerShare` carries the cost of rebalancing a 2x leveraged position, so the 24h window the guidelines ask for is dominated by noise: measured on the same day, one pool annualises to -55% and another to +41%, while the actual share price moved a fraction of a percent. A reader who lands on the card sees a headline rate that describes one day, not the position.

Would you accept `apyBase` computed over the 7-day window for this project (with the 1-day value dropped, or kept only in `apyBase7d`)? Happy to switch it either way, just do not want to deviate from the guidelines without asking.

chadoagent:
Addressed the review note on `apyBase`: the adapter no longer publishes `0` when the historical `pricePerShare` read is unavailable (reverted archive call, or a market younger than the lookback window). Both `apyBase` and `apyBase7d` are now omitted in that case instead of being reported as a measured 0%.

Test run after the change: `Test Suites: 1 passed`, `Tests: 169 passed, 169 total`, 10 pools exported. All four live markets still carry both fields; only a market without history would drop them.

src/adaptors/yield-basis/index.js
```diff
@@ -0,0 +1,351 @@
...
+          apyBase: annualize(pricePerShare, toBigNumber(pps1d[index]), 1),
+        }),
+        ...(toBigNumber(pps7d[index]).gt(0) && {
+          apyBase7d: annualize(pricePerShare, toBigNumber(pps7d[index]), 7),
+        }),
+        ...(pps > 0 && { pricePerShare: pps }),
+        underlyingTokens: [market.asset],
+        token: market.lt,
+        poolMeta: `Unstaked LT, market #${market.id}, Curve pool ${market.cryptopool}`,
+        url: URL,
```
0xkr3p, line 180: i believe this url is incorrect, can we also link to each specific market pls

src/adaptors/yield-basis/index.js
```diff
@@ -0,0 +1,351 @@
...
+        chain: utils.formatChain(CHAIN),
+        project: PROJECT,
+        symbol,
+        tvlUsd: stakedTvlUsd,
+        apyReward,
+        rewardTokens: [ybToken],
+        underlyingTokens: [market.asset],
+        token: market.staker,
+        poolMeta: `Staked gauge, market #${market.id}, Curve pool ${market.cryptopool}`,
+        url: URL,
```
0xkr3p, line 249: same applies here

src/adaptors/yield-basis/index.js
```diff
@@ -0,0 +1,351 @@
...
+      emissions1d,
+      ltSymbols,
+    }),
+  ];
+};
+
+module.exports = {
+  timetravel: false,
+  apy,
+  url: URL,
```
0xkr3p, line 349: and here, this url is not correct, i believe this should https://yieldbasis.com

0xkr3p (commented):
hi @chadoagent, thanks picking this up - the apyBase is definitely spikey, and I think we should instead remove apyBase7d property and assign the value to apyBase to give a more accurate representation

chadoagent:
@0xkr3p thanks for the review, all of it is in:

- `url` is now `https://yieldbasis.com`, and every pool row links to its own market page: `https://yieldbasis.com/earn/<ltAddress>` (the app's `/market/<ltAddress>` route redirects there).
- `apyBase7d` is removed. `apyBase` now carries the 7-day `pricePerShare` change, so the daily spikes are gone. The 1-day slice was dropped entirely rather than left unused.

CI on the latest commit: 169/169 tests, 19 pools.

0xkr3p (approved):
thanks for the fast changes, i made a couple of tweaks to filter markets and reduced the pool meta

PR #3006, opened 2026-09-16
Add Elara Finance yield adapter

vlad-truefi:
## Summary by CodeRabbit

* **New Features**
  * Added coverage for Elara Finance’s Ethereum sElUSD staking pool.
  * Reports annualized APY, TVL, price per share, and the sampled block used for calculations.
  * Preserves negative returns and excludes pools when supply is zero or timing data is invalid.

* **Documentation**
  * Added adapter documentation covering calculation methods, data sources, testing commands, and publishing requirements.


0xkr3p (commented):
thanks for the pr @vlad-truefi, a few things to resolve before we can move forwards:

- lets remove the script, references to the script in index.js and readme
- apy base should be computed over 24h window ideally, if results are spotty then we can use 7d
- use defillama for pricing of ELUSD
- remove the poolmeta

vlad-truefi:
Thanks @0xkr3p! Updated the PR:
- Removed the standalone check script and its references.
- Switched APY to a rolling 7-day window. Recent on-chain samples showed empty 24-hour windows between payouts, so 7 days smooths that timing effect.
- Switched elUSD pricing to DefiLlama.
- Removed poolMeta.

The price API currently returns no data for `ethereum:0x65fb0f9b196d524de0c4f3baf572f0a79eb21194`, which blocks live validation. Could you confirm the correct price identifier or help enable coverage?

We’d also like to support Verified APY for sElUSD. Where can we find the withdrawal-simulation adapter interface or an example integration?

